PR #14, opened 2017-09-07
Fix #10, other changes

### LINKED ISSUE #10: Request validations in campaign controller
We currently assume the input is correct - for example that the application or campaign exists. We need to check these inputs and return appropriate error responses.

martintupy:
Reworked delete campaign method. Reworked exceptions defined in controllers, merged into one class. Added validations to methods in campaign controller (Fix #10). Minor changes, code formatted, removed blank spaces.


powerauth-push-model/src/main/java/io/getlime/push/model/response/DeleteCampaignResponse.java
```diff
@@ -14,20 +14,21 @@
...
+        return deleted;
     }
 
-    public void setId(Long id) {
-        this.id = id;
+    public void setDeleted(boolean deleted) {
+        this.deleted = deleted;
     }
+
+    private boolean deleted;
```
petrdvorak: Please have properties at the beginning of the file.

PR #20, opened 2017-09-13
Implemented client methods (Fix #16), other changes.

### LINKED ISSUE #16: Implement client methods for new server endpoints
Currently, we published new server methods, however we do not have their counterparts on the client side. Therefore, we need to extent the `PushServerClient` to include the new methods.

This task has two parts:

- Implementing the generic `GET` method support - currently, the client only has `POST` support via `sendObjectImpl` method.
- Adding new methods for newly published endpoints.

martintupy:
Modified names, formatted code, added documentation.
Also merged with remote repo.

### DIFF
--- a/powerauth-push-model/src/main/java/io/getlime/push/model/response/DeleteCampaignResponse.java
+++ b/powerauth-push-model/src/main/java/io/getlime/push/model/response/DeleteCampaignResponse.java
@@ -21,8 +21,6 @@ package io.getlime.push.model.response;
  * @author Martin Tupy, martin.tupy.work@gmail.com
  */
 public class DeleteCampaignResponse {
-    private boolean deleted;
-
     public boolean isDeleted() {
         return deleted;
     }
@@ -30,4 +28,7 @@ public class DeleteCampaignResponse {
     public void setDeleted(boolean deleted) {
         this.deleted = deleted;
     }
+
+    private boolean deleted;
+
 }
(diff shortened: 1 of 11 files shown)


powerauth-push-server/src/main/java/io/getlime/push/controller/rest/CampaignController.java
```diff
@@ -229,9 +230,9 @@ public Response addUsersToCampaign(@PathVariable(value = "id") Long campaignId,
      * @param request List of users to remove
      * @return Response status
      */
-    @RequestMapping(value = "{id}/user/remove")
+    @RequestMapping(value = "{id}/user/remove", method = RequestMethod.PUT)
```
petrdvorak: - Use `{id}/user/delete` instead of `{id}/user/remove` to keep the naming consistent.
- Why `PUT` method? Let's use `POST` for deleting...

powerauth-push-model/src/main/java/io/getlime/push/model/response/DeleteCampaignResponse.java
```diff
@@ -21,13 +21,14 @@
...
-    private boolean deleted;
-
     public boolean isDeleted() {
         return deleted;
     }
 
     public void setDeleted(boolean deleted) {
         this.deleted = deleted;
     }
+
```
petrdvorak: Move properties to the beginning of the class.

PR #24, opened 2017-09-20
Changes based on wiki editing.

martintupy:


powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -102,6 +102,18 @@ public void setServiceBaseUrl(String serviceBaseUrl) {
...
     /**
+     * Returns service information
+     *
+     * @return True if service is running.
+     */
+    public boolean getServiceStatus() throws PushServerClientException {
+        TypeReference<ServiceStatusResponse> typeReference = new TypeReference<ServiceStatusResponse>() {
+        };
+        ObjectResponse<?> response = getObjectImpl("/push/service/status", null, typeReference);
+        return response.getStatus().equals(Response.Status.OK);
```
petrdvorak: Return the whole `ServiceStatusResponse` here - there are more information that may be useful to the developer in the response object...

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -152,6 +164,21 @@ public boolean deleteDevice(Long appId, String token) throws PushServerClientExc
...
+     *
+     * @param activationId Identifier of activation
+     * @return True if updating went successful, false otherwise.
+     */
+    public boolean updateDeviceStatus(String activationId) throws PushServerClientException {
+        UpdateDeviceStatusRequest request = new UpdateDeviceStatusRequest();
+        request.setActivationId(activationId);
+        TypeReference<Response> typeReference = new TypeReference<Response>() {
+        };
+        ObjectResponse<?> response = putObjectImpl("/push/device/status/update", request, typeReference);
```
petrdvorak: Device status update must be `POST` - this method is called by the core of PowerAuth Server as an activation status change callback, it cannot be changed to `PUT`.

powerauth-push-docs/sql/mysql/create_push_server_schema.sql
```diff
@@ -12,14 +12,14 @@ CREATE TABLE `push_app_credential` (
...
+  `id` bigint(20) NOT NULL AUTO_INCREMENT,
   `activation_id` varchar(37) DEFAULT NULL,
   `user_id` varchar(255) DEFAULT NULL,
-  `app_id` int(11) DEFAULT NULL,
+  `app_id` bigint(20) DEFAULT NULL,
   `platform` varchar(20) DEFAULT NULL,
   `push_token` varchar(255) DEFAULT NULL,
-  `last_registered` TIMESTAMP NULL,
-  `is_active` int(11) DEFAULT NULL,
+  `timestamp_created` TIMESTAMP NULL,
```
petrdvorak: This is not correct - the timestamp represents "last registration". The mobile device can register multiple times and we are changing this timestamp after each re-registration.

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/DeviceController.java
```diff
@@ -80,7 +80,7 @@ void setClient(PowerAuthServiceClient client) {
             registration.setAppId(appId);
             registration.setPushToken(pushToken);
         }
-        registration.setLastRegistered(new Date());
+        registration.setTimestampCreated(new Date());
```
petrdvorak: The same issue with SQL. It is not our goal to have all timestamps called "timestamp created" - we need to name the timestamp based on what the timestamp represents. This one is timestamp of last device registration and should be named appropriately...

powerauth-push-server/src/main/java/io/getlime/push/repository/model/PushDeviceEntity.java
```diff
@@ -56,8 +56,8 @@
     @Column(name = "push_token", nullable = false)
     private String pushToken;
 
-    @Column(name = "last_registered", nullable = false)
-    private Date lastRegistered;
+    @Column(name = "timestamp_created", nullable = false)
```
petrdvorak: Same as above, the timestamp name must represent actual meaning of the timestamp.

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
petrdvorak: This has to be changed in the controller (in annotation) as well - we have discussed this and the method must be POST despite the method name.

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -125,7 +125,7 @@ else if (platform.equals(PushDeviceEntity.Platform.Android)) {
     }
 
     // Send message to iOS platform
```
petrdvorak: Please write proper JavaDoc comments here - we need to have documentation for parameters. Also, this refactoring does not make much sense - we cannot just introduce methods with 7 parameters, we need functional refactoring...

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -173,7 +173,7 @@ public void operationComplete(Future<PushNotificationResponse<SimpleApnsPushNoti
     }
 
     // Send message to Android platform
-    private void sendMessageToAndroid(FcmClient fcmClient, final Phaser phaser, final PushSendResult result, PushMessage pushMessage, PushDeviceEntity device, final PushMessageEntity sentMessage) {
+    private void sendMessageToAndroid(FcmClient fcmClient, final Phaser phaser, final PushMessageSendResult result, PushMessage pushMessage, PushDeviceEntity device, final PushMessageEntity sentMessage) {
```
petrdvorak: Same as with iOS method above...

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -208,7 +208,7 @@ public void onSuccess(ResponseEntity<String> stringResponseEntity) {
         });
     }
 
-    // If application exists return list of devices related to that app
+    // If user exists return list of devices related to that app
```
petrdvorak: I do not understand purpose of this method - comment is inconsistent with method signature, method returns APNS client.

PR #45, opened 2017-10-06
Updated documentation Fix#33. Added request validators #32. Fix #38 added timestamps to campaign.

### LINKED ISSUE #38: Mark campaign as sent after it is sent
We need two updates:

- mark timestamp in the campaign entity after it is sent
- after the job completes, we should mark campaign as sent

Current status:

screen shot 2017-10-01 at 14 39 46

martintupy:


powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -142,7 +141,7 @@ public boolean createDevice(Long appId, String token, MobilePlatform platform, S
         request.setActivationId(activationId);
         TypeReference<Response> typeReference = new TypeReference<Response>() {
         };
-        ObjectResponse<?> response = postObjectImpl("/push/device/create", new ObjectRequest<>(request), typeReference);
+        ObjectResponse<?> response = postObjectImpl("/push/device/create", request, typeReference);
```
petrdvorak: Oh boy... Why do you make changes like this? :) Why is this supposed to continue to work, after you arbitrarily change the request object type? The change is not correct, and I have noticed that even the new methods that you implemented (delete, update status, ...) have the same issue and use incorrect request object...

This method (and all other) must always call:

```java
new ObjectRequest<>(request)
```

Try to figure out why is that...

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -435,7 +455,13 @@ public boolean sendCampaign(Long campaignId) throws PushServerClientException {
...
     }
 
-    private <T> ObjectResponse<T> checkStatus(TypeReference typeReference, HttpResponse response) throws IOException, PushServerClientException {
+    /**
+     * Checks if format of http response is valid
+     *
+     * @param typeReference reference on type of response body from which map into JSON
+     * @param response prepared http response
+     */
+    private <T> ObjectResponse<T> checkHttpResponseFormat(TypeReference typeReference, HttpResponse response) throws IOException, PushServerClientException {
```
petrdvorak: Why did you rename this method and added incorrect comment about what it does? This method explicitly checks HTTP status code value (it checks nothing else) and serializes response object accordingly. Name `checkStatus` was OK, maybe `checkHttpStatus` would be better, but these renames are unnecessary...

powerauth-push-model/src/main/java/io/getlime/push/model/validator/SendPushMessageBatchRequestValidator.java
```diff
@@ -0,0 +1,46 @@
...
+    public static String validate(SendPushMessageBatchRequest pushMessageBatch) {
+        if (pushMessageBatch == null) {
+            return "Empty request";
+        } else if (pushMessageBatch.getAppId() == null) {
+            return "Empty appId";
+        } else if (pushMessageBatch.getBatch().size() > 20) {
+            return "Too many messages in batch - do no send more than 20 messages at once to avoid server congestion.";
+        } else {
+            for (PushMessage pushMessage : pushMessageBatch.getBatch()) {
+                if (pushMessage == null) {
```
petrdvorak: This can never happen, you ask for push message object from collection.

On the other hand, you should check for:

```java
if (pushMessageBatch.getBatch() == null) {
    return "No messages in batch";
}
```

... in the beginning of this method.

PR #76, opened 2017-12-15
Fix #39 entities - equals hashcode

### LINKED ISSUE #39: JPA entities are missing equals() and hashCode()
It seems that relying on default implementation of equals() and hashCode() is not a good idea for JPA entities, see:

https://vladmihalcea.com/2016/06/06/how-to-implement-equals-and-hashcode-using-the-jpa-entity-identifier

I think we should look into this, equals() and hashCode() implementations are omitted in Push Server.

martintupy:


petrdvorak:
Looks OK, I will wait with merging this one until I thoroughly test this in order to make sure the change didn't break anything.

powerauth-push-server/src/main/java/io/getlime/push/repository/model/PushMessageEntity.java
```diff
@@ -160,4 +160,24 @@ public Status getStatus() {
...
+        if (!getDeviceId().equals(that.getDeviceId())) return false;
+        if (!getMessageBody().equals(that.getMessageBody())) return false;
+        return getTimestampCreated().equals(that.getTimestampCreated());
+    }
+
+    @Override
+    public int hashCode() {
+        int result = getDeviceId().hashCode();
+        result = 31 * result + getMessageBody().hashCode();
+        result = 31 * result + getTimestampCreated().hashCode();
```
romanstrobl: There's a small bug on this row - contents of result are overwritten.

powerauth-push-server/src/main/java/io/getlime/push/repository/model/PushCampaignUserEntity.java
```diff
@@ -86,4 +86,24 @@ public Date getTimestampCreated() {
...
+        if (!getCampaignId().equals(that.getCampaignId())) return false;
+        if (!getUserId().equals(that.getUserId())) return false;
+        return getAppId().equals(that.getAppId());
+    }
+
+    @Override
+    public int hashCode() {
+        int result = getCampaignId().hashCode();
+        result = 31 * result + getUserId().hashCode();
+        result = 31 * result + getAppId().hashCode();
```
romanstrobl: There's a small bug on this row - contents of result are overwritten.

powerauth-push-server/src/main/java/io/getlime/push/repository/model/PushMessageEntity.java
halendew: I can't see what exactly is going to be overwritten, true is that attribute values should be used instead of getters.

romanstrobl: Sorry, I overlooked it, it is OK.

powerauth-push-server/src/main/java/io/getlime/push/repository/model/PushCampaignUserEntity.java
romanstrobl: Not a problem, either.

PR #85, opened 2018-02-05
Fix #84: Server ends up hanging in case of invalid error

### LINKED ISSUE #84: Server ends up hanging in case of invalid error

petrdvorak:


petrdvorak:
@halendew The error is logged in the beginning of the `if ... else ...` statement [one level above](https://github.com/lime-company/powerauth-push-server/blob/dff05e719d8ce350dbd3f6f99fa572a056971a55/powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java#L325), there is no additional contextual info we need to log.

PR #91, opened 2018-03-27
Prepare release 0.18.0

petrdvorak:


powerauth-push-server/pom.xml
```diff
@@ -123,7 +123,20 @@
 			<scope>test</scope>
 		</dependency>
 
-	</dependencies>
+		<!--swagger-->
```
romanstrobl: Please check the formatting, it seems to be inconsistent.

petrdvorak: I will try to fix it.

petrdvorak:
Fixed formatting.

PR #102, opened 2018-05-14
Refactor the class PushMessageSenderService by extracting a smaller class

A4Vision:


powerauth-push-server/src/main/java/io/getlime/push/service/EnhancedPushConfiguration.java
```diff
@@ -0,0 +1,197 @@
...
+import java.io.IOException;
+import java.net.InetSocketAddress;
+import java.security.InvalidKeyException;
+import java.security.NoSuchAlgorithmException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class EnhancedPushConfiguration {
```
petrdvorak: I like where this change is going but I'm a bit unsure about the class naming. This class does push message sending and as such, it is rather `*Worker` than `*Configuration`. Let's call the class `PushSendingWorker` instead.

romanstrobl: I do not like the class name EnhancedPushConfiguration, perhaps we can use a better word in the class name than Enhanced? Furthermore, the class contains methods which send messages, so it should not be called "Configuration", too. I believe a "Configuration" class should only contain logic related to actual configuration, not message delivery.

powerauth-push-server/src/main/java/io/getlime/push/service/EnhancedPushConfiguration.java
```diff
@@ -0,0 +1,197 @@
...
+    PushServiceConfiguration pushServiceConfiguration;
```
romanstrobl: The field is not declared as private, it should not be accessible from other classes.

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -91,7 +75,7 @@ public PushMessageSenderService(AppCredentialsRepository appCredentialsRepositor
         this.appCredentialsRepository = appCredentialsRepository;
         this.pushDeviceRepository = pushDeviceRepository;
         this.pushMessageDAO = pushMessageDAO;
-        this.pushServiceConfiguration = pushServiceConfiguration;
+        this.enhancedPushConfiguration = new EnhancedPushConfiguration(pushServiceConfiguration);
```
petrdvorak: We should use autowiring here, not using `new`...

powerauth-push-server/src/main/java/io/getlime/push/service/EnhancedPushConfiguration.java
petrdvorak: Agreed.

powerauth-push-server/src/main/java/io/getlime/push/service/EnhancedPushConfiguration.java
petrdvorak: ... thank you, I overlooked this one.

petrdvorak:
Hello @A4Vision,

thank you very much for the pull request! :) We see where this is going and agree that the code for `PushMessageSenderService` could use some cleanup. I have reviewed your pull request with my colleague @romanstrobl and we found couple minor places where the code could be improved.

Can you please have a look at our comments? If not, we will take your PR as an inspiration and incorporate similar change ourselves.

With kind regards,

Petr Dvorak

A4Vision:
Thanks for your comments.

- Renamed class to `PushSendingWorker`.
- Changed field `pushServiceConfiguration` to private.
- Changed constructor to auto wiring instead of calling `new `
- Defined the `PushSendingWorker` bean as a "Service" component.

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -254,7 +236,7 @@ public void sendCampaignMessage(Long appId, String platform, String token, PushM
      * @throws PushServerException In case any issue happens while sending the push message. Detailed information about
      * the error can be found in exception message.
      */
-    public void sendCampaignMessage(final Long appId, String platform, final String token, PushMessageBody pushMessageBody, PushMessageAttributes attributes, String userId, Long deviceId, String activationId) throws PushServerException {
+    private void sendCampaignMessage(final Long appId, String platform, final String token, PushMessageBody pushMessageBody, PushMessageAttributes attributes, String userId, Long deviceId, String activationId) throws PushServerException {
```
petrdvorak: This method is supposed to be public. While the compiler probably complains with an info message that "The access can be private" (since this method is not used at the moment with controller code), this is an Java code API method extending the `sendCampaignMessage()` method that is just above with possibility to use `attributes`.

A4Vision:
>  iOS sending code should be moved as well
- That's a good point - moved it. @petrdvorak 

> This method is supposed to be public
- Changed to public. @romanstrobl

petrdvorak:
Hello @A4Vision - thank you very much for the pull request, now it is ready for merge! :) Before we can do this, we need you to review and agree to our CLA. Please review it (it is reasonably short and readable) here and if everything is OK for you, please confirm the agreement: https://goo.gl/forms/2FQ5hH0uxz7YVphG2

petrdvorak:
@A4Vision Thank you very much for the CLA approval and for the contribution! :)

PR #107, opened 2018-05-18
Fix #101: Add better logging to push client code

### LINKED ISSUE #101: Add better logging to push client code
Currently, we do throw exceptions, but out log says nothing about the actual issue...

petrdvorak:
Added logging methods to the client calls.
Used model validators on the client side.
Fixed problem with BuildProperties autowiring in unit tests.

### DIFF
--- a/powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClientException.java
+++ b/powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClientException.java
@@ -29,7 +29,13 @@ public class PushServerClientException extends Exception {
 
     private Error error;
 
+    public PushServerClientException(String message) {
+        super(message);
+        this.error = new PushServerClientError(message);
+    }
+
     public PushServerClientException(Error error) {
+        super(error.getMessage());
         this.error = error;
     }
 
--- a/powerauth-push-model/src/main/java/io/getlime/push/model/entity/ListOfUsers.java
+++ b/powerauth-push-model/src/main/java/io/getlime/push/model/entity/ListOfUsers.java
@@ -17,13 +17,20 @@
 package io.getlime.push.model.entity;
 
 import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Request used for adding collection of users to campaign
  *
  * @author Martin Tupy, martin.tupy.work@gmail.com
  */
-
 public class ListOfUsers extends ArrayList<String> {
+
+    public ListOfUsers() {
+    }
+
+    public ListOfUsers(List<String> users) {
+        addAll(users);
+    }
 
 }
--- a/powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
+++ b/powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
@@ -36,6 +36,7 @@ import io.getlime.push.model.entity.PushMessageBody;
 import io.getlime.push.model.entity.PushMessageSendResult;
 import io.getlime.push.model.request.*;
 import io.getlime.push.model.response.*;
+import io.getlime.push.model.validator.*;
 
 import java.io.IOException;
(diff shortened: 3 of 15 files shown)


powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClientException.java
```diff
@@ -29,7 +29,13 @@
 
     private Error error;
 
+    public PushServerClientException(String message) {
+        super(message);
+        this.error = new PushServerClientError(message);
+    }
+
     public PushServerClientException(Error error) {
+        super(error.getMessage());
```
romanstrobl: When error is null, NPE will be thrown.

powerauth-push-model/src/main/java/io/getlime/push/model/entity/ListOfUsers.java
```diff
@@ -17,13 +17,20 @@
...
 import java.util.ArrayList;
+import java.util.List;
 
 /**
  * Request used for adding collection of users to campaign
  *
  * @author Martin Tupy, martin.tupy.work@gmail.com
  */
-
 public class ListOfUsers extends ArrayList<String> {
```
romanstrobl: I am not sure what to think about this code - why are we extending an ArrayList? It does not seem necessary.

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClientException.java
petrdvorak: ... good point.

powerauth-push-model/src/main/java/io/getlime/push/model/entity/ListOfUsers.java
petrdvorak: I have no strong opinion here - why is extending `ArrayList` bad? The `ListOfUsers` class represents RESTful API entity that is a collection of user identifiers. Having own class for that might have better semantics than just generic `ArrayList<String>`...

petrdvorak: ... plus it is a good way to reference specific type, for example in `ListOfUsersFromCampaignResponse` ([here](https://github.com/lime-company/powerauth-push-server/blob/cfbf5c76ae85f5f0e49b85aa722c968d1bfd8155/powerauth-push-model/src/main/java/io/getlime/push/model/response/ListOfUsersFromCampaignResponse.java)). We also have some responses that are lists themselves, such as `ListOfCampaignsResponse` ([here](https://github.com/lime-company/powerauth-push-server/blob/cfbf5c76ae85f5f0e49b85aa722c968d1bfd8155/powerauth-push-model/src/main/java/io/getlime/push/model/response/ListOfCampaignsResponse.java)).

romanstrobl: Let's search the internets for answers...

https://www.ibm.com/developerworks/java/library/j-jtp02216/#2.1

https://softwareengineering.stackexchange.com/questions/246277/good-or-bad-practice-to-mask-java-collections-with-meaningful-class-names

petrdvorak: Ok, so based on the articles, we can probably leave it as is.

- There is often an argument "use composition instead". This is correct for examples like `class School extends ArrayList<Student>`, where students are actually one of many properties of school and using array list seems like an abuse.
- Also, there is an argument about object comparison and lookup. This is not relevant for us, use use these classes as model class for RESTful APIs and we need them to only bind with the underlying JSON.
- Arguments about performance will not be the issue anymore, it is not 2006.

romanstrobl: Ok

PR #119, opened 2018-06-01
Migration to WebClient for FCM

### LINKED ISSUE #103: Consider a thread-pooling TaskExecutor
We currently use default `SimpleAsyncTaskExecutor`, we should consider using other implementation.

### LINKED ISSUE #106: Remove deprecated AsyncRestTemplate
Spring Boot 2.0 deprecated `AsyncRestTemplate`, we should replace it with some updated mechanism.

romanstrobl:
* Fix #106: Remove deprecated AsyncRestTemplate
* Fix #103: Consider a thread-pooling TaskExecutor


romanstrobl:
Please leave this PR open for now, we should use subscribe() on WebClient to avoid uncessary blocking in reactive code. I will update the code.

powerauth-push-server/src/main/java/io/getlime/push/service/fcm/FcmClient.java
```diff
@@ -16,82 +16,109 @@
...
+import reactor.ipc.netty.options.ClientProxyOptions;
+
+import java.util.function.Consumer;
 
 /**
  * FCM server client
  *
  * @author Petr Dvorak, petr@lime-company.eu
  */
+@Service
```
petrdvorak: Just a quick note before it is finished: this is completely wrong. We need to construct multiple instance of `FcmClient`, one for each application that is configured in the system (with credentials that are specific for that app). This is why `prepareFcmClient` uses `new` keyword to create a new instance...

romanstrobl: I see, ok, I will revert the change and handle autowiring of configuration otherwise.

romanstrobl:
After learning more about reactive programming in Spring it is clear that we should not use @Async and thread pooled task executor, because it is the "old" way of doing things. Instead we should apply so called backpressure using a Subscriber. 

I believe I wasn't getting a HTTP response in WebClient without blocking previously because the doOnSuccess() method only receives a signal and doesn't cause any backpressure. This is why I had to originally use @Async and block() to obtain a response. 

Currently the subscribe() method waits for a HTTP response and Consumers are called when Mono<FcmSendResponse> or Throwable is extracted from HTTP response. 

This removes the need for @Async completely.

romanstrobl:
This PR is ready for review. I will work on performance tests now to verify the scalability of FCM HTTP requests.

PR #157, opened 2018-08-16
Add admin REST API and migrate web interface to the REST API

romanstrobl:
This PR adds REST API for administration in Push Server.

I haven't removed the web interface in Push Server yet, it was migrated to REST API instead. The web controller code will be reused in PA universal admin, so it should not be an worthless effort. We can remove the web interface completely once PA universal admin is available.

@hvge I am adding you as reviewer as agreed with @petrdvorak.


romanstrobl:
I raised issues for the enhancements:
https://github.com/wultra/powerauth-push-server/issues/158
https://github.com/wultra/powerauth-push-server/issues/159

PR #160, opened 2018-08-27
Fix #116: Rewrite FCM support to use new FCM API v1

### LINKED ISSUE #116: Rewrite FCM support to use new FCM API v1
Currently, we use "Legacy HTTP API" for interacting with FCM. We should rewrite the client to use the newer `v1` API...

romanstrobl:


powerauth-push-server/src/main/java/io/getlime/push/service/PushSendingWorker.java
```diff
@@ -106,78 +112,112 @@ FcmClient prepareFcmClient(String serverKey) {
...
+            for (Map.Entry<String, Object> entry : extras.entrySet()) {
+                data.put(entry.getKey(), entry.getValue().toString());
+            }
+        }
+
+        AndroidConfig.Builder androidConfigBuilder = AndroidConfig.builder()
+                .setCollapseKey(pushMessageBody.getCollapseKey());
+
+        if (pushServiceConfiguration.isFcmDataNotificationOnly()) { // notification only through data map
+            String notificationString = pushMessageBody.getTitle()
```
petrdvorak: Let's use multiple keys: `_notification.title`, ... etc.

petrdvorak: And let's do an experiment by still using key `_notification` with the `Object` payload.

powerauth-push-server/src/main/resources/application.properties
```diff
@@ -14,6 +14,7 @@ spring.datasource.password=
 spring.datasource.driver-class-name=com.mysql.jdbc.Driver
 spring.jpa.properties.hibernate.connection.characterEncoding=utf8
 spring.jpa.properties.hibernate.connection.useUnicode=true
+spring.jpa.properties.hibernate.dialect=org.hibernate.dialect.MySQLDialect
```
petrdvorak: Remove

powerauth-push-server/src/main/resources/application.properties
```diff
@@ -74,4 +75,6 @@ powerauth.push.service.apns.connect.timeout=5000
 spring.jpa.hibernate.use-new-id-generator-mappings=false
 
 # Disable open session in view to avoid startup warning of Spring boot
-spring.jpa.open-in-view=false
\ No newline at end of file
+spring.jpa.open-in-view=false
+
+logging.level.reactor.ipc.netty=debug
```
petrdvorak: Remove

romanstrobl:
I fixed all the change requests. The only remaining item is to verify that we are ok with minor change of serialization of data-only notifications.

Data-only notifications in 2018.06 as seen in Android device:
image

Current data-only notifications:
image

The difference is in ordering attributes and ignoring null values.

PR #188, opened 2019-01-09
Fix #186: Message persistence should be optional

### LINKED ISSUE #186: Message persistence should be optional
Push server stores delivered messages and updates their status based on result of communication with FCM and APNS. We should make the message persistence optional to speed up message delivery and avoid clogging the database in case the Push Server deployment does not require storage of sent messages and their statuses.

We should use a configuration property to disable message persistence.

romanstrobl:


powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -97,7 +100,12 @@ public PushMessageSendResult sendPushMessage(final Long appId, List<PushMessage>
 
             // Iterate over all devices for given user
             for (final PushDeviceRegistrationEntity device : devices) {
-                final PushMessageEntity pushMessageObject = pushMessageDAO.storePushMessageObject(pushMessage.getBody(), pushMessage.getAttributes(), pushMessage.getUserId(), pushMessage.getActivationId(), device.getId());
+                final PushMessageEntity pushMessageObject;
+                if (configuration.isMessageStorageEnabled()) {
```
petrdvorak: Just for the sake of slightly better code clarity, I would probably isolate this to a private method as well, right above that `updateStatusAndPersist ` method.

powerauth-push-server/src/main/resources/application.properties
```diff
@@ -44,6 +44,9 @@ powerauth.push.service.applicationEnvironment=
 # PowerAuth 2.0 Push Campaign Setup
 powerauth.push.service.campaign.batchSize=100000
 
+# Whether persistent storing of sent messages is enabled
+powerauth.push.service.message.storage.enabled=true
```
petrdvorak: Let's set this to false by default - I believe that storing the push messages proved useless so far and only seem to hurt performance. For migrating customers, we need to explicitly mention this behaviour change...

romanstrobl: Ok, see: https://github.com/wultra/powerauth-push-server/wiki/PowerAuth-Push-Server-0.21.0#storing-of-sent-push-messages-disabled-by-default

petrdvorak:
Looks great now. :)

PR #208, opened 2019-02-06
Fix #207: Handle case when HTTP response contains no body for a non-200 status code

### LINKED ISSUE #207: Handle case when HTTP response contains no body for a non-200 status code
In case a non-200 status code is returned and the HTTP response contains no body, the following error occurs:

```
io.getlime.push.client.PushServerClientException: com.fasterxml.jackson.databind.exc.MismatchedInputException: No content to map due to end-of-input
 at [Source: (ByteArrayInputStream); line: 1, column: 0]
	at io.getlime.push.client.PushServerClient.postObjectImpl(PushServerClient.java:696) ~[powerauth-push-client-0.21.0.jar:0.21.0]
	at io.getlime.push.client.PushServerClient.postObjectImpl(PushServerClient.java:667) ~[powerauth-push-client-0.21.0.jar:0.21.0]
	at io.getlime.push.client.PushServerClient.createDevice(PushServerClient.java:163) ~[powerauth-push-client-0.21.0.jar:0.21.0]
```

We should handle this case and avoid parsing empty HTTP response. The error occurs when push server is not running (404 status code is returned).

romanstrobl:


powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -758,10 +760,21 @@ public Response removeAndroid(Long id) throws PushServerClientException {
...
+            String responseBody = null;
+            try (Scanner scanner = new Scanner(response.getRawBody(), StandardCharsets.UTF_8.name()).useDelimiter("\\A")) {
+                if (scanner.hasNext()) {
+                    responseBody = scanner.next();
+                }
+            }
+            if (responseBody != null && !responseBody.isEmpty()) {
+                // Response body contains data, return Exception with status code and error response
+                com.fasterxml.jackson.databind.ObjectMapper mapper = new com.fasterxml.jackson.databind.ObjectMapper();
+                ErrorResponse errorResponse = mapper.readValue(responseBody, ErrorResponse.class);
```
petrdvorak: ... this can probably fail in case there is some error HTML code, we should check for valid JSON. This will also handle the empty response (not valid JSON), so the error handling can be unified.

romanstrobl:
All serialization errors should be now resolved using a try-catch block.

PR #271, opened 2019-10-28
Fix #270: Update dependencies

### LINKED ISSUE #270: Update dependencies

petrdvorak:


romanstrobl:
We should probably test that push messages are delivered after these changes.

petrdvorak:
> We should probably test that push messages are delivered after these changes.

Sure - let's build it and prepare a deploy to WUAT...

PR #278, opened 2019-11-14
Fix #255: Revise the build process to avoid bundling undesired files

### LINKED ISSUE #255: Revise the build process to avoid bundling undesired files
We can be J-Rebels and bundle every XML file with our packages, or we can clean the process up...

romanstrobl:


romanstrobl:
Hm I see, perhaps a better solution is to use Jenkins to build our artifacts, then we will not have such an issue because JRebel will not be installed. I think I can close this pull request without merging because the issue will go away once we move Jenkins jobs to new machine and build wars using Jenkins.

PR #279, opened 2019-11-14
Fix #269: Proxy authentication support for FCM

### LINKED ISSUE #269: Proxy authentication support for FCM
`GoogleCredential` class uses OAuth 2.0 protocol  to obtain access tokens, however the default Google HTTP client based on java.net HTTP client does not support proxy authentication:

https://github.com/googleapis/google-http-java-client/issues/190

We can workaround this issue by using the `java.net.Authenticator` for release `0.22.x`, given that the Google `NetHttpTransport` class does not allow setting proxy authentication. This cannot be considered a proper solution though, just a workaround of missing functionality in Google HTTP client library.

To fix the issue properly in the future we can consider switching to `google-http-client-apache-v2` which should support Proxy authentication.

romanstrobl:
I think we can use the implementation with `Authenticator`, although it is not ideal, it is only used when proxy authentication is set and will complete the proxy support in Push Server.


romanstrobl:
I tested it on a proxy server which required proxy authentication.

pmagam-tibco:
Hi
I'm using ccproxy.exe as proxy server permitting only user/password account.With below code I'm getting error as "Unable to tunnel through proxy. Proxy returns HTTP/1.1 407 Proxy Authentication Required".Is it how to use Authenticator for FCM oauth proxy authentication?

Authenticator.setDefault(new Authenticator() {
            @Override
            protected PasswordAuthentication getPasswordAuthentication() {
                return new PasswordAuthentication("username", "password".toCharArray());
            }
        });

		// Set up HTTP transport with proxy
		
		HttpTransport httpTransport = null;				
				NetHttpTransport.Builder netHttpTransportBuilder = new NetHttpTransport.Builder();
				netHttpTransportBuilder.setProxy(new Proxy(Proxy.Type.HTTP, new InetSocketAddress(httpClient.getHostConfiguration().getProxyHost(), httpClient.getHostConfiguration().getProxyPort())));
				httpTransport = netHttpTransportBuilder.build();
		
		GoogleCredential googleCredential = null;
		 googleCredential = GoogleCredential.fromStream(
				new FileInputStream(filePath), httpTransport, com.google.api.client.googleapis.util.Utils.getDefaultJsonFactory())
				.createScoped(Arrays.asList(arr));

petrdvorak:
@pmagam-tibco We cannot help with this. As at https://github.com/googleapis/google-http-java-client

PR #280, opened 2019-11-20
Update device registration in push server

### LINKED ISSUE #252: Add "associated activations" to registration request
As an optional feature, we could subscribe the token to multiple "associated" activation IDs. This feature must be opt-in as it finds a use case only in rare situations (master-child multi-banking model) and it introduces potential security issues (legitimate user can create a signed request and subscribe to someone else's notifications in case activation ID is revealed).

### LINKED ISSUE #262: Make sure push token is unique for given activation ID
Currently, it seems that both Apple and Google issue new push token despite the fact the old didn't expire yet. We don't know why this happens but we need to handle this better.

romanstrobl:
Following issues should be fixed:
- Fix #252: Add "associated activations" to registration request
- Fix #262: Make sure push token is unique for given activation ID


powerauth-push-server/src/main/java/io/getlime/push/controller/rest/PushDeviceController.java
```diff
@@ -87,26 +87,99 @@ void setClient(PowerAuthServiceClient client) {
         String pushToken = requestedObject.getToken();
         String platform = requestedObject.getPlatform();
         String activationId = requestedObject.getActivationId();
-        PushDeviceRegistrationEntity device = pushDeviceRepository.findFirstByAppIdAndPushToken(appId, pushToken);
-        if (device == null) {
+        // Make sure push token is unique for activation ID, in case activation ID is specified in request. See: https://github.com/wultra/powerauth-push-server/issues/262
+        // Both Apple and Google issue new push token despite the fact the old one didn't expire yet. The old push tokens need to be deleted.
+        if (activationId != null) {
+            pushDeviceRepository.deleteByAppIdAndActivationId(appId, activationId);
```
petrdvorak: Activation ID is almost always present - except for some corner cases where device registration occurs for non-active devices. My concern with this logic is that if we do this, we will never find the device later (will be likely deleted), and we will be creating a new record for every registration. As a result, we will run out of `rid` index values soon...

Basically, as long as there is only one record for a given activation ID, we should be updating it with a new push token.

Note that we also need to do the opposite logic - in case a user removes activation from a device and creates a new one for given user, the activation ID will change and push token will remain the same.

petrdvorak: We should create a unique index on `(push_token)`, `(activation_id)` and `(push_token, activation_id)` columns, to force data consistency on DB level.

romanstrobl: The first and third indeces are in contradiction, aren't they?

petrdvorak: @romanstrobl Nope. That is the point and this is why it is difficult.

Case 1: You use an active app. APNs decides to give you a new push token.
=> You need to make sure that there is only one push token for given activation ID.

Case 2: John uses an active app, APNs decides to give him a push token X. John unpairs the app and gives phone to Jane. Push token is the same, activation changes.
=> We need to make sure only to associate the push token to the last activation.

petrdvorak: (Truth is that activation ID is unique only in case it is non-null.)

romanstrobl: I mean we were discussing the use case when we had two rows with different activation IDs but same push token. This use case would not be possible due to first index, although the third index allows it.

petrdvorak: Ah... gotcha... for multi-activation registration, "push token" alone does not have to be unique...

romanstrobl: Let's discuss these issues in person, I will prepare list of use cases and some proposal in meanwhile, because we need to support multiple use cases.

romanstrobl:
Implementation and documentation is ready, I will work on:
- tests, there are many branches
- prepare SQL update queries for removing duplicates before unique indexes are applied

PR #281, opened 2019-12-07
Prepare release 0.23.0

romanstrobl:
This pull request contains all changes between releases `0.22.0` and `0.23.0`.


docs/PowerAuth-Push-Server-0.23.0.md
```diff
@@ -0,0 +1,25 @@
...
+
+The DDL script for MySQL:
+```sql
+CREATE UNIQUE INDEX `push_device_activation_token` ON `push_device_registration`(`activation_id`, `push_token`);
+
+DROP INDEX `push_device_activation` ON `push_device_registration`;
+CREATE UNIQUE INDEX `push_device_activation` ON `push_device_registration`(`activation_id`);
+```
+
+In case either of the index updates fails, delete existing duplicate rows. Rows with newest timestamp_last_registered should be preserved.
```
petrdvorak: I think we are missing the mention of `powerauth.push.service.registration.multipleActivations.enabled` here, and maybe generally in the docs.

romanstrobl: Ok, I will add it.

romanstrobl:
I have updated the documentation: https://github.com/wultra/powerauth-push-server/pull/281/commits/cb86f0445ac66abae2bd5e02cf2e811d82925ac3. Please verify it and let me know if we can proceed with the release. We can also prepare the SQL query to remove duplicit registrations (basically rows with expired `pushtoken` values which should no longer be present in the database), but this documentation does not block the release.

PR #287, opened 2020-03-19
Fix #286: Remove static Unirest initialization and update to Kong version

### LINKED ISSUE #286: Remove static Unirest initialization and update to Kong version
The Unirest instances are configured using a static call to the main Unirest class - it is not wise to configure it inside the library. Instead, it is smart to advice additional configuration to the library user.

petrdvorak:


pom.xml
```diff
@@ -86,7 +86,7 @@
         <powerauth.version>0.24.0-SNAPSHOT</powerauth.version>
         <pushy.version>0.13.10</pushy.version>
         <swagger.version>2.9.2</swagger.version>
-        <unirest.version>1.4.9</unirest.version>
+        <unirest.version>3.4.01</unirest.version>
```
saalistaja: Could we use higher version here, 3.4.04?
Or ideally the latest one https://mvnrepository.com/artifact/com.konghq/unirest-java

petrdvorak: Good point... latest is 3.6.01 - IMO it is typo in their docs here: http://kong.github.io/unirest-java/

petrdvorak: Fixed - I updated all libs...

saalistaja: ok!

:) that's a classic case with the documentation and reality

PR #288, opened 2020-03-23
Integrate changes from develop

romanstrobl:
The H2 DB downgrade needs to be integrated into the release.

I also integrated the Unirest object mapper and separated the version for Push client.


romanstrobl:
I found out what was the problem, the version of PowerAuth and Push Client is not the same, so I had to fix it and then your Unirest object mapper commit could be integrated.

PR #313, opened 2020-07-07
Fix #312: Consider migration to WebClient

### LINKED ISSUE #312: Consider migration to WebClient
The majority of PowerAuth backend stack is either using or migrating to WebClient. We should consider using WebClient in `powerauth-push-client` to unify the HTTP client libraries and provide similar configuration options.

romanstrobl:
This should be the last client to be migrated to WebClient.

### DIFF
--- a/powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
+++ b/powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
@@ -754,7 +754,7 @@ public class PushServerClient {
      */
     @SuppressWarnings("unchecked")
     private <T> T checkHttpStatus(ParameterizedTypeReference<? extends Response> typeReference, ClientResponse response) throws IOException, PushServerClientException {
-        if (!response.statusCode().isError()) {
+        if (response.statusCode().is2xxSuccessful()) {
             return (T) response.bodyToMono(typeReference).block();
         } else {
             try {
--- a/powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
+++ b/powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
@@ -17,9 +17,7 @@
 package io.getlime.push.client;
 
 import com.fasterxml.jackson.core.JsonParseException;
-import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.JsonMappingException;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.io.BaseEncoding;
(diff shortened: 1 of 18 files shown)


powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -625,17 +626,14 @@ public Response removeAndroid(Long id) throws PushServerClientException {
...
-                    .header("Accept", "application/json")
-                    .header("Content-Type", "application/json")
-                    .queryString(params)
-                    .asString();
-            return checkHttpStatus(typeReference, response);
-        } catch (UnirestException e) {
-            logger.warn(e.getMessage(), e);
-            throw new PushServerClientException(e, new Error("PUSH_SERVER_CLIENT_ERROR", "Network communication has failed."));
+            ClientResponse response = webClient.get()
+                    .uri(uriBuilder -> uriBuilder.path(url).queryParams(params).build())
```
petrdvorak, line 632: Just to check... Does this work even in case the "base URL" contains context?

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -749,17 +753,17 @@ public Response removeAndroid(Long id) throws PushServerClientException {
      * @throws IOException In case JSON processing fails.
      */
     @SuppressWarnings("unchecked")
-    private <T> T checkHttpStatus(TypeReference<? extends Response> typeReference, HttpResponse<String> response) throws IOException, PushServerClientException {
-        if (response.getStatus() == 200) {
-            return (T) mapper.readValue(response.getBody(), typeReference);
+    private <T> T checkHttpStatus(ParameterizedTypeReference<? extends Response> typeReference, ClientResponse response) throws IOException, PushServerClientException {
+        if (!response.statusCode().isError()) {
```
petrdvorak, line 757: Could we make this condition positive (maybe `is2xxSuccessful()`) or flip the branches? The `!` with `isError()` feels like a "double negation"...

romanstrobl:
I will test it.

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
romanstrobl: We can test the context easily by migrating WebFlow to the new client. I will prepare this change in a branch.

romanstrobl: I have confirmed that the constructed URL is correct, WebFlow uses context `/powerauth-push-client` and the new client works fine.

PR #322, opened 2020-08-24
Fix #321: Error after migrating to Instant

### LINKED ISSUE #321: Error after migrating to Instant
The following error occurs due to migration to recent `Instant`:
```
i.g.p.r.serialization.JSONSerialization  : Cannot construct instance of `java.time.Instant` (no Creators, like default construct, exist): cannot deserialize from Object value (no delegate- or property-based Creator)
```

romanstrobl:


petrdvorak (approved):
Looks OK

PR #351, opened 2020-10-21
Fix #347: Review the client rewrite to WebClient

### LINKED ISSUE #347: Review the client rewrite to WebClient
The `PushServerClient.checkHttpStatus` contains a redundant `throws IOException`. When you remove it, more errors pop out. Exceptions in `putObjectImpl` and `postObjectImpl` are no longer thrown and we do not need to catch them.

petrdvorak:
Still several things unresolved:

- What about JSON parsing errors?
- Where is HTTP status error handling?
- Does the updated method signature work out?


romanstrobl:
Let's migrate the following classes to the `rest-client-base` library which provides a common implementation of the REST client:
- `FcmClient`
- `PushServerClient`

romanstrobl (commented):
I have migrated the REST clients to rest-client-base. Let's review the merged code changes.

powerauth-push-server/src/main/java/io/getlime/push/service/fcm/FcmClient.java
```diff
@@ -107,29 +106,22 @@ public void setProxySettings(String proxyHost, int proxyPort, String proxyUserna
...
-                                builder.username(proxyUsername);
-                                builder.password(s -> proxyPassword);
-                            }
-                            builder.build();
-                        });
-                    }
-                    return tcpClient;
-                });
-        webClient = WebClient.builder().clientConnector(new ReactorClientHttpConnector(httpClient)).build();
+    public void initializeWebClient() throws PushServerException {
```
petrdvorak, line 111: We should rename method and fix the JavaDoc.

romanstrobl:
Looks good now.

PR #356, opened 2020-11-16
Fix #355: Remove reactor-spring dependency

### LINKED ISSUE #355: Remove reactor-spring dependency
The `reactor-spring` dependency is no longer needed and it causes reported CVE issues.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #358, opened 2020-11-19
Fix #357: Remove obsolete resources

### LINKED ISSUE #357: Remove obsolete resources
The push server still contains obsolete resources which should be removed because the web interface is no longer present.

romanstrobl:


romanstrobl:
Ah, I see, the resources are used by the home page. We'll need to update jQuery then.

PR #366, opened 2020-12-09
Fix #365: Use object methods from REST client

### LINKED ISSUE #365: Use object methods from REST client
Now that all request/response types are covered in `rest-model-base` we can migrate to *object methods instead of using generic types. This simplifies the code except for 1 case where a generic method is used (for PowerAuth status update).

romanstrobl:
Migration from generic methods call to *object method calls and simplification of FCM message serialization.

### DIFF
--- a/powerauth-push-server/src/main/java/io/getlime/push/service/fcm/model/MessageWithValidation.java
+++ b/powerauth-push-server/src/main/java/io/getlime/push/service/fcm/model/MessageWithValidation.java
@@ -15,6 +15,7 @@
  */
 package io.getlime.push.service.fcm.model;
 
+import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.firebase.messaging.Message;
 
 /**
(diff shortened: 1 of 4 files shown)


powerauth-push-server/src/main/java/io/getlime/push/service/fcm/model/MessageWithValidation.java
```diff
@@ -0,0 +1,77 @@
...
+
+/**
+ * Model class for FCM requests with message and validation mode.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+public class MessageWithValidation {
+
+    private Message message;
+    private boolean validate_only;
```
petrdvorak, line 28: Let's use JSON annotation to assure the underscore-case, let's use cammel-case in Java.

petrdvorak (approved):
Looks OK, I have a minor style comment...

powerauth-push-server/src/main/java/io/getlime/push/service/fcm/model/MessageWithValidation.java
romanstrobl: Ok, fixed.

PR #367, opened 2020-12-10
Revert serialization change

romanstrobl:
Revert serialization change and fix error with blocking call when extracting the response.


petrdvorak (approved):
Looks ok

PR #390, opened 2021-03-20
Fix #388: Use docucheck metadata in RESTful docs

### LINKED ISSUE #388: Use docucheck metadata in RESTful docs

petrdvorak:


docs/Push-Server-API.md
```diff
@@ -1,4 +1,5 @@
 # Push Server API
+<!-- template api -->
```
hvge, line 2: I think that this line has no effect.

hvge (approved):
Looks OK

docs/Push-Server-API.md
petrdvorak: It does - it selects the correct page template in the documentation portal.

PR #400, opened 2021-05-05
Fix #399: Do not include title, message and sound for silent push notifications

### LINKED ISSUE #399: Do not include title, message and sound for silent push notifications

petrdvorak:


romanstrobl (approved):
Looks good

PR #405, opened 2021-05-24
Fix #404: Migrate to WebClient version 5.3, update versions and dependencies

### LINKED ISSUE #404: Migrate to WebClient version 5.3
Migration of FCM callbacks requires:
- use `ResponseEntity<String>` instead of `ClientResponse` due to migration of `exchange` -> `exchangeToMono`
- error handling uses `RestClientException` instead of `WebClientException` due to changes in error handling

romanstrobl:


powerauth-push-server/src/main/java/io/getlime/push/service/PushSendingWorker.java
```diff
@@ -149,13 +151,10 @@ void sendMessageToAndroid(final FcmClient fcmClient, final PushMessageBody pushM
             callback.didFinishSendingMessage(PushSendingCallback.Result.FAILED);
         };
 
-        // Callback when FCM request succeeds
```
petrdvorak, line 152: 🤔 Hmm... why was this here...

petrdvorak (approved):
Looks OK

powerauth-push-server/src/main/java/io/getlime/push/service/PushSendingWorker.java
romanstrobl: Yes, however, note that the original Rest client code with `ClientResponse` was a bit more generic - there was full access to the response including HTTP headers, etc. Subscribing on `exchangeToMono` provides `ResponseEntity<T>` or exception where we can gather all the details. This handling should be hopefully fine for our use cases.

PR #410, opened 2021-05-31
Fix #409: Remove dependency on sun.security.provider.X509Factory

### LINKED ISSUE #409: Remove dependency on sun.security.provider.X509Factory
We only use it because of the well-known certificate file prefixes.

petrdvorak:


romanstrobl (approved):
Easy one.

Issue #420, opened 2021-07-27
Log details of error for FCM messages


petrdvorak:
Currently, we have some `invalid-argument` error and we are blind about the reason.

PR #422, opened 2021-07-27
Backport #420 to releases/1.1.x

petrdvorak:


romanstrobl (approved):
I think that this change may cause duplicate logging for the known error codes. Wouldn't it be better to add this log message into the `default` case in the switch only? At any case I approve the pull request, because it improves the logging, I am just concerned about logging the known errors twice.

petrdvorak:
@romanstrobl The goal of the change is to add error details for all possible responses, not just for default behavior. The issue is that the current logging does not allow investigating existing exceptions, since the error response has more details.

romanstrobl:
I see, then the change is fine.

PR #423, opened 2021-07-27
Fix #419: Add support for localizable arguments

### LINKED ISSUE #419: Add support for localizable arguments
We should add support for:

### Android

- title_loc_key
- title_loc_args
- body_loc_key
- body_loc_args

### iOS

- title-loc-key
- title-loc-args
- loc-key
- loc-args

Note: This also requires update in the docs for payload mapping.

petrdvorak:


powerauth-push-server/src/main/java/io/getlime/push/service/PushSendingWorker.java
```diff
@@ -250,7 +251,11 @@ private Message buildAndroidMessage(final PushMessageBody pushMessageBody, final
         final AndroidNotification notification = AndroidNotification.builder()
                 .setPriority(deliveryPriority)
                 .setTitle(pushMessageBody.getTitle())
+                .setTitleLocalizationKey(pushMessageBody.getTitleLocKey())
+                .addAllTitleLocalizationArgs(Arrays.asList(pushMessageBody.getTitleLocArgs()))
```
romanstrobl, line 255: We should probably add a null pointer check for `pushMessageBody.getTitleLocArgs()`, otherwise I think that an NPE will be thrown in `Arrays.asList()`.

powerauth-push-server/src/main/java/io/getlime/push/service/PushSendingWorker.java
```diff
@@ -250,7 +251,11 @@ private Message buildAndroidMessage(final PushMessageBody pushMessageBody, final
...
                 .setBody(pushMessageBody.getBody())
+                .setBodyLocalizationKey(pushMessageBody.getBodyLocKey())
+                .addAllBodyLocalizationArgs(Arrays.asList(pushMessageBody.getBodyLocArgs()))
```
romanstrobl, line 258: dtto

romanstrobl (changes requested):
Looks good, but let's handle the null cases.

romanstrobl (approved):
Looks good now

PR #431, opened 2021-08-06
Fix #429: Document priority and localization

### LINKED ISSUE #429: Document priority and localization
We added new push payload attributes for setting notification priority and localization attributes. We need to update documentation with these attributes.

petrdvorak:


romanstrobl (approved):
Looks good

PR #435, opened 2021-08-11
Fix #434: Migrate to v7 error handling for FCM

### LINKED ISSUE #434: Migrate to v7 error handling for FCM
The `firebase-admin-java` introduced new error handling for FCM in v7: https://github.com/firebase/firebase-admin-java/issues/354

Push server needs to be migrated to the new error handling, the change seems to be incompatible in the REST API as far as I can tell due to changes of the `errorCode` constants.

romanstrobl:
This change should fix the issues with resolving FCM errors. I will test it tomorrow, which is why this PR is opened as a draft right now.


romanstrobl:
I went through the changes in FCM Admin Java SDK. Most of the error codes are indeed compatible, because the all caps version of error code is used for mapping to the low caps error code. However, the compatibility is not 100% -- e.g. `THIRD_PARTY_AUTH_ERROR` is now present in the `MessagingErrorCode` enumeration, but it was missing in the original error code map, so this error code is not handled properly in current release of Push Server.

petrdvorak (approved):
Looks OK

PR #436, opened 2021-08-16
Fix #430: Use blob for storing iOS private key on PostgreSQL

### LINKED ISSUE #430: Use blob for storing iOS private key on PostgreSQL
We now use VARCHAR(255), which is almost too short...

romanstrobl:


petrdvorak (approved):
Looks good

PR #437, opened 2021-08-17
Fix #352: Allow the per-app configuration of APNS environment

### LINKED ISSUE #352: Allow the per-app configuration of APNS environment

petrdvorak:


docs/PowerAuth-Push-Server-1.2.0.md
```diff
@@ -1,6 +1,34 @@
...
+#### PostgreSQL
+
+```sql
+ALTER TABLE push_app_credentials ADD ios_environment VARCHAR(32) NULL;
+```
+
+#### Oracle
+
+```sql
+ALTER TABLE push_app_credentials ADD ios_environment VARCHAR2(32) NULL;
```
romanstrobl, line 22: I think it is better to use `VARCHAR2(32 CHAR)`. We should probably use it in all VARCHARs for Oracle in Push Server, because by default the limit is specified as number of bytes, not as number of characters.

romanstrobl (approved):
I noticed one minor issue related to VARCHAR2 in Oracle. Feel free to raise it as a separate issue because the issue is present in other columns, too.

docs/PowerAuth-Push-Server-1.2.0.md
petrdvorak: Good catch! I fixed this in b3632f8, including one `VARCHAR` vs. `VARCHAR2` issue.

PR #442, opened 2021-09-01
Fix #441: Switch default database connection to PostgreSQL

### LINKED ISSUE #441: Switch default database connection to PostgreSQL
Let's switch the default database connection defined in `application.properties` to PostgreSQL.

romanstrobl:


petrdvorak (approved):
Looks good

PR #448, opened 2021-12-15
Fix #447: Exclude log4j-api dependency

### LINKED ISSUE #447: Exclude log4j-api dependency
Although PowerAuth Push Server is not vulnerable in default configuration with the `Log4Shell` attack, we should exclude the `log4j-api` dependency to avoid raising alerts in various vulnerability scans.

romanstrobl:
This change should resolve any future questions about the Log4J dependencies included in Spring logging.


petrdvorak (approved):
Looks OK

PR #456, opened 2022-01-24
Fix #455: Messages not delivered to APNs in version 1.1.1

### LINKED ISSUE #455: Messages not delivered to APNs in version 1.1.1
In a scenario where `titleLocKey` is missing the push message is not delivered, although APNs response confirms successful delivery. 

The problem seems to be that we are using an older version of the pushy library `0.14.2` where behaviour of localized messages is different from version `0.15.0`. Thus the issue is not reproducible in version `1.2.0` of push server.

romanstrobl:


petrdvorak (approved):
Looks OK

PR #463, opened 2022-05-05
Fix #461: Add support for logging using correlation ID with MDC

### LINKED ISSUE #461: Add support for logging using correlation ID with MDC
We should add a configurable option to include correlation ID from HTTP headers in server logs.

Following changes are required:
- add configuration property to enable logging with correlation ID, e.g. `*.correlation-header.enabled=true`
- add configuration property for correlation ID header name, e.g. `*.correlation-header.name=Correlation-ID`
- add the correlation ID value into MDC context
- document how to set up logging with correlation headers using Spring configuration properties

romanstrobl:


petrdvorak (approved):
Looks OK. I do not like the trick with MDC though, this idea od "Let's have Singleton bean and access a threat local context ad-hoc in it" feels a bit difficult to read with confidence.

PR #472, opened 2022-06-06
Fix #468: Use development APNs as default

### LINKED ISSUE #468: Use development APNs as default
Most of the initial deployments are with non-proction environments and hence having this as a default seems practical. When changing the value, we must clearly mention it in the documentation so that the customers who upgrade do not rely on default value.

petrdvorak:


romanstrobl (approved):
This change is pretty useful.

PR #483, opened 2022-06-20
Fix #473: Add information about database dialect configuration into migration guide

### LINKED ISSUE #473: Add information about database dialect configuration into migration guide
In current code the default DB dialect is specified, otherwise Spring application would not start.

We need to document the configuration of database dialect in migration guide for clients who use Oracle, otherwise the default PostgreSQL dialect is used which causes invalid SQL queries to be generated by Hibernate when using Oracle.

romanstrobl:


banterCZ (approved):

docs/PowerAuth-Push-Server-1.3.0.md
```diff
@@ -83,4 +83,19 @@ Also, in several API calls, the `id` variable was renamed to `appId`, to maintai
...
\ No newline at end of file
+  - `DELETE /admin/app/android/remove`
+
+## Database Dialect Configuration
+
+The latest release of PowerAuth requires configuration of database dialect.
+
+The dialect is specified using following configuration property:
+```properties
+spring.jpa.database-platform=org.hibernate.dialect.PostgreSQLDialect
```
banterCZ, line 94: We are using a deprecated value in the example. Could we change it?

romanstrobl: Sure, good idea.

PR #487, opened 2022-08-25
Fix #486: Implement inbox for messages

### LINKED ISSUE #486: Implement inbox for messages

petrdvorak:
What to focus at:

- The indexes in the DB - are they optimal?
- The SQL script changes on Oracle.
- Client is currently untested, I will do the test integration to powerauth-cloud and enrollment-server components.


docs/PowerAuth-Push-Server-1.4.0.md
```diff
@@ -0,0 +1,64 @@
...
+CREATE INDEX push_inbox_user_read ON push_inbox (user_id, timestamp_created, read);
+```
+
+#### Oracle
+
+```sql
+CREATE TABLE PUSH_INBOX (
+  ID VARCHAR2(37 CHAR) PRIMARY KEY,
+  USER_ID VARCHAR2(255 CHAR) NOT NULL,
+  SUBJECT BLOB NOT NULL,
```
romanstrobl, line 37: I believe `CLOB` is the equivalent of `TEXT`. It is better for storing text and searching though it than binary data.

docs/PowerAuth-Push-Server-1.4.0.md
```diff
@@ -0,0 +1,64 @@
...
+  BODY BLOB NOT NULL,
```
romanstrobl, line 38: dtto

docs/PowerAuth-Push-Server-1.4.0.md
```diff
@@ -0,0 +1,64 @@
...
+  READ NUMBER(1) DEFAULT false,
```
romanstrobl, line 39: Should probably be `NOT NULL`.

docs/PowerAuth-Push-Server-1.4.0.md
```diff
@@ -0,0 +1,64 @@
...
+
+#### PostgreSQL
+
+```sql
+CREATE TABLE push_inbox (
+    id VARCHAR(37) CONSTRAINT push_inbox_pk PRIMARY KEY,
+    user_id VARCHAR(255) NOT NULL,
+    subject TEXT NOT NULL,
+    body TEXT NOT NULL,
+    read BOOLEAN DEFAULT false,
```
romanstrobl, line 22: Should probably be `NOT NULL`.

docs/PowerAuth-Push-Server-1.4.0.md
```diff
@@ -0,0 +1,64 @@
...
+
+#### MySQL
+
+```sql
+CREATE TABLE `push_inbox` (
+  `id` VARCHAR(37),
+  `user_id` VARCHAR(255) NOT NULL,
+  `subject` TEXT NOT NULL,
+  `body` TEXT NOT NULL,
+  `read` BOOLEAN DEFAULT false,
```
romanstrobl, line 56: Should probably be `NOT NULL`.

powerauth-push-model/src/main/java/io/getlime/push/model/response/GetInboxMessageCountResponse.java
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Response object for count of messages in inbox.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Data
+@AllArgsConstructor
+public class GetInboxMessageCountResponse {
+
+    long count;
```
romanstrobl, line 31: Missing `private` modifier.

docs/PowerAuth-Push-Server-1.4.0.md
```diff
@@ -0,0 +1,64 @@
...
+    user_id VARCHAR(255) NOT NULL,
+    subject TEXT NOT NULL,
+    body TEXT NOT NULL,
+    read BOOLEAN DEFAULT false,
+    timestamp_created TIMESTAMP NOT NULL,
+    timestamp_read TIMESTAMP
+);
+
+CREATE INDEX push_inbox_user ON push_inbox (user_id);
+CREATE INDEX push_inbox_user_read ON push_inbox (user_id, timestamp_created, read);
```
romanstrobl, line 28: I am missing following index:
- `user_id`, `read`

I think the index on (`user_id`, `timestamp_created`, `read`) is not very beneficial, because `timestamp_created` is not used after WHERE, it is only used in ORDER BY. The DB can sort results which are not too big (up to tens of thousands) fast without index. 

We certainly need a dedicated index for the select with `user_id` and `read` in WHERE condition.

powerauth-push-server/src/main/java/io/getlime/push/service/InboxService.java
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    @Transactional
+    public GetInboxMessageDetailResponse postMessage(String userId, CreateInboxMessageRequest request) {
+        final InboxMessageEntity messageEntity = inboxMessageConverter.convert(UUID.randomUUID(), userId, request, new Date());
+        final InboxMessageEntity savedMessageEntity = inboxRepository.save(messageEntity);
+        return inboxMessageConverter.convertResponse(savedMessageEntity);
+    }
+
+    public ListOfInboxMessages fetchMessageListForUser(String userId, boolean onlyUnread, Pageable pageable) {
```
romanstrobl, line 61: I would add `@Transactional(readOnly=true)` or a comment why transactions are not required for reading.

powerauth-push-server/src/main/java/io/getlime/push/service/InboxService.java
```diff
@@ -0,0 +1,99 @@
...
+        final List<InboxMessageEntity> messageEntities;
+        if (onlyUnread) {
+            messageEntities = inboxRepository.findAllByUserIdAndReadOrderByTimestampCreatedDesc(userId, false, pageable);
+        } else {
+            messageEntities = inboxRepository.findAllByUserIdOrderByTimestampCreatedDesc(userId, pageable);
+        }
+        return inboxMessageConverter.convert(messageEntities, pageable.getPageNumber(), pageable.getPageSize());
+    }
+
+    public GetInboxMessageDetailResponse fetchMessageDetail(String userId, String id) throws InboxMessageNotFoundException {
```
romanstrobl, line 71: I would add `@Transactional(readOnly=true)` or a comment why transactions are not required for reading.

powerauth-push-server/src/main/java/io/getlime/push/service/InboxService.java
```diff
@@ -0,0 +1,99 @@
...
+        final Optional<InboxMessageEntity> messageEntity = inboxRepository.findFirstByIdAndUserId(id, userId);
+        if (!messageEntity.isPresent()) {
+            throw new InboxMessageNotFoundException("Unable to fetch message: " + id + ", for user: " + userId + ".");
+        }
+        return inboxMessageConverter.convertResponse(messageEntity.get());
+    }
+
+    public GetInboxMessageCountResponse fetchMessageCountForUser(String userId) {
```
romanstrobl, line 79: I would add `@Transactional(readOnly=true)` or a comment why transactions are not required for reading.

powerauth-push-server/src/main/java/io/getlime/push/errorhandling/DefaultExceptionHandler.java
```diff
@@ -103,4 +104,19 @@ public ErrorResponse handlePowerAuthClientException(Exception e) {
         logger.error(e.getMessage(), e);
         return new ErrorResponse(WebServiceError.Code.ERROR_PA_SERVER_COMM, e);
     }
+
+    /**
+     * Handle exceptions related to inbox.
+     * @param e {@link InboxMessageNotFoundException}
+     * @return Error response.
+     */
+    @ResponseStatus(HttpStatus.NOT_FOUND)  // 404
```
romanstrobl, line 113: I am not sure about the `404`, it will be also returned when the application is not deployed. We can keep it, let's just be aware of this.

powerauth-push-model/src/main/java/io/getlime/push/model/request/CreateInboxMessageRequest.java
```diff
@@ -0,0 +1,43 @@
...
+@Data
+public class CreateInboxMessageRequest {
+
+    @NotNull
+    @Size(min = 8, max = 255)
+    @Schema(type = "string", example = "Example subject")
+    private String subject;
+
+    @NotNull
+    @Size(min = 8)
```
romanstrobl, line 39: I think we still need a `max` parameter, but very large to avoid DOS'ing the DB.

docs/PowerAuth-Push-Server-1.4.0.md
```diff
@@ -0,0 +1,64 @@
...
+);
+
+CREATE INDEX PUSH_INBOX_USER ON push_inbox(USER_ID);
+CREATE INDEX PUSH_INBOX_USER_READ ON push_inbox (USER_ID, TIMESTAMP_CREATED, READ);
+```
+
+#### MySQL
+
+```sql
+CREATE TABLE `push_inbox` (
```
romanstrobl, line 51: I am not sure why we keep using backticks for MySQL. There are no special characters in table or column names...

powerauth-push-model/src/main/java/io/getlime/push/model/request/CreateInboxMessageRequest.java
```diff
@@ -0,0 +1,43 @@
...
+
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+
+/**
+ * Request object for the message.
+ *
+ * @author Petr Dvorak, petr@wultra.com
+ */
+@Data
```
romanstrobl, line 30: We should probably exclude `body` from `toString`.

romanstrobl (approved):
In general this is a nice new functionality. 

I am a bit concerned that there is no message expiration, so the inbox will always grow over time in size, but that can be implemented later.

I found some issue worth reviewing.

powerauth-push-model/src/main/java/io/getlime/push/model/response/GetInboxMessageCountResponse.java
petrdvorak: ... I was tired. :D

docs/PowerAuth-Push-Server-1.4.0.md
petrdvorak: Good point, thank you! 👍

powerauth-push-server/src/main/java/io/getlime/push/service/InboxService.java
petrdvorak: I added the annotation.

powerauth-push-server/src/main/java/io/getlime/push/errorhandling/DefaultExceptionHandler.java
petrdvorak: I will change this to 400...

powerauth-push-model/src/main/java/io/getlime/push/model/request/CreateInboxMessageRequest.java
petrdvorak: I will restrict this by 65536.

docs/PowerAuth-Push-Server-1.4.0.md
petrdvorak: Consistency with the rest of our MySQL documentation in the repo. :-/

PR #490, opened 2022-09-06
Fix #489: Fix minor tasks related to inbox

### LINKED ISSUE #489: Fix minor tasks related to inbox

petrdvorak:


romanstrobl (changes requested):
I think that the current naming of database columns is inconsistent with the rest of the stack. The database ID should be probably called `id` and the external ID should be called `inbox_id` or `inbox_reference` or anything else than `id` which I think is reserved for DB primary keys.

romanstrobl (approved):
DDL now corresponds to other tables, we can merge this pull request.

PR #494, opened 2022-10-14
Fix #493: Incorrect column type in SQL scripts

### LINKED ISSUE #493: Incorrect column type in SQL scripts

petrdvorak:


romanstrobl (approved):
Looks OK.

PR #497, opened 2022-10-17
Fix #496: Remove build time dependency of power-auth-push-server on running environment

### LINKED ISSUE #496: Remove build time dependency of power-auth-push-server on running environment
remove dependency on powerauth-server in powerauth-push-server junit tests - probably by mocking the dependency

romanstrobl:


powerauth-push-server/src/test/java/io/getlime/push/controller/MockPowerAuthController.java
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    @PostMapping("application/create")
+    public ObjectResponse<CreateApplicationResponse> createApplication() {
+        final CreateApplicationResponse response = new CreateApplicationResponse();
+        return new ObjectResponse<>(response);
+    }
+
+    @PostMapping("application/detail")
+    public ObjectResponse<GetApplicationDetailResponse> getApplicationDetail() {
```
banterCZ, line 63: I know, it is only a test, but better not to swallow the exception. But what about
```suggestion
    public ObjectResponse<GetApplicationDetailResponse> getApplicationDetail() throw CryptoProviderException {
```

powerauth-push-server/src/test/java/io/getlime/push/controller/MockPowerAuthController.java
```diff
@@ -0,0 +1,138 @@
...
+        version.setSupported(true);
+        version.setApplicationKey("key");
+        version.setApplicationSecret("secret");
+        response.getVersions().add(version);
+        try {
+            final KeyPair keyPair = new KeyGenerator().generateKeyPair();
+            final KeyConvertor keyConvertor = new KeyConvertor();
+            response.setMasterPublicKey(BaseEncoding.base64().encode(keyConvertor.convertPublicKeyToBytes(keyPair.getPublic())));
+        } catch (CryptoProviderException e) {
+            // Ignore error in mock endpoint
```
banterCZ, line 77: Or at least
```suggestion
            logger.error("Crypto provider error", e);
```

banterCZ (approved):

powerauth-push-server/src/test/java/io/getlime/push/controller/MockPowerAuthController.java
romanstrobl: We don't want to declare an exception that can never happen.

powerauth-push-server/src/test/java/io/getlime/push/controller/MockPowerAuthController.java
romanstrobl: Yes, logging is fine, but again, the exception cannot happen.

PR #502, opened 2022-11-10
Fix #501: Add application ID parameter to Inbox REST API requests

### LINKED ISSUE #501: Add application ID parameter to Inbox REST API requests
We need to differentiate user IDs by application IDs, so that when there are multiple users with the same user ID the message is not sent to incorrect user. 

This change has impact on REST API whenever the user ID parameter is used. We need to add another mandatory `@RequestParam` called `appId` to differentiate users across applications.

This change requires following changes in code:
- `InboxMessageEntity` - added parameter `appId`, new column `app_id`
- `InboxService` - in method `postMessage` added parameter `appId`,  the record is checked for validity against table `push_app_credentials`
- `InboxService` - methods `fetchMessageListForUser`, `fetchMessageDetail`, `fetchMessageCountForUser` and `readMessage` use the `applicationId` as additional filter

Sample `postMessage`:
POST /inbox/user123?appId=abc123

Note, that the `appId` corresponds to column `push_app_credentials.app_id`, not `push_app_credentials.id` or `pa_application.id`.

No access controls are applied when using the `app_id` when posting messages, the only check is that the `app_id` exists in database.

We also need to update indexes for queries which include the new `app_id` column.

romanstrobl:
Pending tasks:
- [x] testing
- [x] documentation update


romanstrobl:
The pull request is ready for review.

powerauth-push-server/src/main/java/io/getlime/push/repository/InboxRepository.java
```diff
@@ -35,34 +35,38 @@ public interface InboxRepository extends PagingAndSortingRepository<InboxMessage
...
     /**
      * Find first message with given ID and user ID.
-     * @param id Message ID.
+     * @param inboxId Message ID.
      * @param userId User ID.
+     * @param appId Application ID.
      * @return First message matching ID and user ID.
      */
-    Optional<InboxMessageEntity> findFirstByIdAndUserId(String id, String userId);
+    Optional<InboxMessageEntity> findFirstByInboxIdAndUserIdAndAppId(String inboxId, String userId, String appId);
```
banterCZ, line 61: Just consider _parameter object_ pattern, three parameters in a row of the same type are error prone.

banterCZ (approved):
Looks good.

powerauth-push-server/src/main/java/io/getlime/push/repository/InboxRepository.java
romanstrobl: Exactly.

PR #507, opened 2022-11-14
Fix #506: Use queryParams when calling REST client with appId parameter

### LINKED ISSUE #506: User queryParams when calling REST client with appId parameter
We should use the REST client methods with queryParams instead of specifying the parameters in URL directly.

romanstrobl:


powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -846,6 +872,26 @@ private <T> T putImpl(String url, Object request, ParameterizedTypeReference<T>
...
+     * @param typeReference type reference
+     * @return Object obtained after processing the response JSON.
+     * @throws PushServerClientException In case of network, response / JSON processing, or other IO error.
+     */
+    private <T> T putImpl(String url, Object request, MultiValueMap<String, String> queryParams, MultiValueMap<String, String> headers, ParameterizedTypeReference<T> typeReference) throws PushServerClientException {
+        try {
+            return restClient.put(url, request, queryParams, headers, typeReference).getBody();
+        } catch (RestClientException ex) {
+            logger.warn(ex.getMessage(), ex);
+            throw new PushServerClientException(ex, new Error("PUSH_SERVER_CLIENT_ERROR", "HTTP POST request failed."));
```
banterCZ, line 891: 🤔 I am glad that we propagate the cause, but isn't it log twice at the end?

banterCZ (approved):

PR #518, opened 2022-12-06
Fix #517: Fix client call for reading a single inbox message

### LINKED ISSUE #517: Fix client call for reading a single inbox message

petrdvorak:


hvge (approved):
Looks OK

PR #521, opened 2022-12-09
Fix #522: Update Spring Boot to 2.6.14

### LINKED ISSUE #522: Update Spring Boot to 2.6.14

### LINKED ISSUE #791: Update Wultra dependencies

banterCZ:


romanstrobl (approved):
I am not sure it is a good idea to remove Jackson version, because there are usually many vulnerabilities in the library which force us to increase the version independent on Spring Boot versioning. I guess we can bring it back if needed.

banterCZ:
> I am not sure it is a good idea to remove Jackson version, because there are usually many vulnerabilities in the library which force us to increase the version independent on Spring Boot versioning. I guess we can bring it back if needed.

Good point but I have double checked that we are on the right version. This declaration could keep us behind. It is driven by Spring Boot now. May be overridden (just maven property, dependency declaration already inherits the property from the parent) if reported.

PR #527, opened 2022-12-15
Fix #526: Coverity - Unused private method at PushServerClient

### LINKED ISSUE #526: Coverity - Unused private method at PushServerClient

banterCZ:


romanstrobl (approved):
Good catch.

PR #543, opened 2022-12-28
Fix #542: Return more complete detail of inbox message

### LINKED ISSUE #542: Return more complete detail of inbox message

petrdvorak:


powerauth-push-model/src/main/java/io/getlime/push/model/request/CreateInboxMessageRequest.java
```diff
@@ -42,4 +47,8 @@ public class CreateInboxMessageRequest {
     @ToString.Exclude
     private String body;
 
+    @NotNull
+    @NotEmpty
+    private List<@NonNull @Pattern(regexp="^[a-zA-Z0-9_-]{3,255}$") String> applications = new ArrayList<>();
```
romanstrobl, line 52: Are you sure you want to use `lombok.NonNull` annotation in the list?

powerauth-push-server/src/main/java/io/getlime/push/repository/InboxRepository.java
```diff
@@ -37,47 +38,47 @@ public interface InboxRepository extends PagingAndSortingRepository<InboxMessage
...
+    long countAllByUserIdAndApplicationsContainingAndRead(String userId, AppCredentialsEntity app, boolean read);
 
     /**
      * Mark all user messages as read.
      * @param userId User ID.
-     * @param appId Application ID.
+     * @param app Application.
      */
-    @Query("UPDATE InboxMessageEntity i SET i.read = true WHERE i.userId = :userId AND i.appId = :appId AND i.read = false")
+    @Query("UPDATE InboxMessageEntity i SET i.read = true WHERE i.userId = :userId AND :app member i.applications AND i.read = false")
```
romanstrobl, line 80: JPAQL provides the `MEMBER OF` operator, does `MEMBER` operator really work?

romanstrobl (changes requested):
I found two issues in the code which should be resolved.

docs/sql/mysql/create_push_server_schema.sql
```diff
@@ -87,5 +92,7 @@ CREATE INDEX `push_campaign_user_campaign` ON `push_campaign_user`(`campaign_id`
 CREATE INDEX `push_campaign_user_detail` ON `push_campaign_user`(`user_id`);
 
 CREATE INDEX `push_inbox_id` ON `push_inbox` (`inbox_id`);
-CREATE INDEX `push_inbox_user_app` ON `push_inbox` (`user_id`, `app_id`);
+CREATE INDEX `push_inbox_user_app` ON `push_inbox` (`user_id`);
```
romanstrobl, line 95: There is not `app` in the `push_inbox_user_app` index, is the index name correct?

docs/sql/mysql/create_push_server_schema.sql
```diff
@@ -87,5 +92,7 @@ CREATE INDEX `push_campaign_user_campaign` ON `push_campaign_user`(`campaign_id`
...
 CREATE INDEX `push_inbox_user_read` ON `push_inbox` (`user_id`, `read`);
+
+CREATE INDEX `push_inbox_app_inbox` ON `push_inbox_app` (`inbox_id`, `app_credentials_id`);
```
romanstrobl, line 98: This index seems a bit strange, it should be unique, we probably do not want to allow duplicates. Probably a good candidate for a primary key.

docs/sql/mysql/create_push_server_schema.sql
```diff
@@ -68,6 +67,12 @@ CREATE TABLE `push_inbox` (
   PRIMARY KEY (`id`)
 ) ENGINE=InnoDB AUTO_INCREMENT=1 CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;
 
+-- Create table for assignment of inbox messages to apps
+CREATE TABLE `push_inbox_app` (
```
romanstrobl, line 71: The table has no primary key. Usually database tables do have a primary key, unless it is just a dump of data.

romanstrobl (changes requested):
I also noticed that the new DB table `push_inbox_app` is missing a primary key, which is a bit of an antipattern in database design for structured data.

powerauth-push-model/src/main/java/io/getlime/push/model/request/CreateInboxMessageRequest.java
petrdvorak: good catch

powerauth-push-server/src/main/java/io/getlime/push/repository/InboxRepository.java
petrdvorak: Strangely, this seems to work. I will change it to `MEMBER OF`.

docs/sql/mysql/create_push_server_schema.sql
petrdvorak: Thank you for catching this, issue for MySQL and Oracle. PostgreSQL was correct.

docs/sql/mysql/create_push_server_schema.sql
petrdvorak: Absolutely, I will fix this.

docs/sql/mysql/create_push_server_schema.sql
petrdvorak: Agreed.

romanstrobl (approved):
Fix approved.

PR #547, opened 2023-01-03
Fix #546: Move user ID from path variable to query parameter and request body

### LINKED ISSUE #546: Move user ID from path variable to query parameter and request body
Since the `userID` parameter is selected by the calling party (according to the user naming), we might avoid putting the user ID directly in the path variable. We should unify the inbox controller behavior with other controller classes.

petrdvorak:


powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -622,21 +615,17 @@ public ObjectResponse<GetInboxMessageDetailResponse> postMessage(String userId,
...
     public PagedResponse<ListOfInboxMessages> fetchMessageListForUser(String userId, List<String> applications, boolean onlyUnread, Integer page, Integer size) throws PushServerClientException {
-        try {
-            final String userIdSanitized = URLEncoder.encode(String.valueOf(userId), "UTF-8");
-            final MultiValueMap<String, String> params = buildPages(page, size);
-            params.add("applications", String.join(",", applications));
-            params.add("onlyUnread", onlyUnread ? "true" : "false");
+        final MultiValueMap<String, String> params = buildPages(page, size);
+        params.add("userId", userId);
+        params.add("applications", String.join(",", applications));
+        params.add("onlyUnread", onlyUnread ? "true" : "false");
```
banterCZ, line 621:
```suggestion
        params.add("onlyUnread", Boolean.toString(onlyUnread));
```

banterCZ (approved):
Looks good.

PR #553, opened 2023-01-03
Fix #548: Inbox: validate application in message list

### LINKED ISSUE #548: Inbox: validate application in message list
The message list endpoint does not validate whether applications is configured in push server and returns empty message list for application which is not configured.

petrdvorak:


romanstrobl (approved):
OK for the first version of Inbox.

PR #565, opened 2023-01-04
Fix #508: Add instructions for migrating application ID on Oracle

### LINKED ISSUE #508: Add instructions for migrating application ID on Oracle
The following sample script does not work on Oracle:
https://github.com/wultra/powerauth-push-server/blob/develop/docs/PowerAuth-Push-Server-1.3.0.md#migrating-application-id

Oracle does not support `IF NOT EXISTS` and `UPDATE ... SET ... FROM ... WHERE` syntax.

romanstrobl:


petrdvorak (approved):
Looks OK

banterCZ (approved):
Looks good.

PR #568, opened 2023-01-11
Fix #567: Invalid database mapping

### LINKED ISSUE #567: Invalid database mapping
We currently map columns the other way around due to incorrect JPA mapping.

Screenshot 2023-01-11 at 19 54 42

Fixing this makes version 1.4.0 unfit for deployment.

petrdvorak:


romanstrobl (approved):
Good catch!

PR #573, opened 2023-01-12
Fix #570: Add plain text summary and mime-type to inbox message model

### LINKED ISSUE #570: Add plain text summary and mime-type to inbox message model
Currently, we cannot work with HTML content properly, and we can only receive plain text properly. This could be too restrictive to some of the customers.

Screenshot

We need to add:

- `summary` attribute - a plain text summary of the message with at most 256 characters (to be used in push and list of messages)
- `type` attribute - a mime type attribute `application/text` and `applications/html` (maybe just `text` and `html`, to keep it simple).

petrdvorak:


banterCZ (approved):
Looks good. The Coverity scan also passed.

PR #586, opened 2023-03-16
Backport fix #462: Add support for asynchronous requests

### LINKED ISSUE #462: Add support for asynchronous requests
Sometimes, the client may not care about the response details we provide. To allow faster synchronous processing on client side, we should allow asynchronous messages.

- The client would add `"mode": "asynchronous"` to the request (default would be `"mode": "synchronous"`).
- The server would return a mode in the response and if the mode is `asynchronous`, there would be no details of sent messages.

petrdvorak:


banterCZ (approved):

romanstrobl (approved):
Looks OK.

PR #601, opened 2023-03-31
Fix #600: Turn off auto-commit

### LINKED ISSUE #600: Turn off auto-commit
See https://vladmihalcea.com/why-you-should-always-use-hibernate-connection-provider_disables_autocommit-for-resource-local-jpa-transactions/

banterCZ:


powerauth-push-server/src/main/java/io/getlime/push/configuration/BatchSendingConfiguration.java
```diff
@@ -176,6 +177,8 @@ public void initialize() {
 
     private PlatformTransactionManager createTransactionManager() {
         final JpaTransactionManager transactionManager = new JpaTransactionManager();
+        transactionManager.setDataSource(dataSource);
```
banterCZ, line 180: See https://github.com/spring-projects/spring-batch/issues/4051#issuecomment-1066630060

PR #608, opened 2023-04-04
Fix #566: App ID migration drops unique index on app ID

### LINKED ISSUE #566: App ID migration drops unique index on app ID
Due to rename of the column and drop of the renamed column the index on `app_id` column is also dropped and it is not reintroduced during the migration. See: https://github.com/wultra/powerauth-push-server/blob/develop/docs/PowerAuth-Push-Server-1.3.0.md#migrating-application-id

banterCZ:


romanstrobl (approved):
Good catch!

PR #613, opened 2023-04-06
Fix #612: Use @Transactional annotation from Spring

### LINKED ISSUE #612: Use @Transactional annotation from Spring
We should use `@Transactional` annotation from Spring as in the rest of the Wultra stack.

banterCZ:


romanstrobl (approved):
Looks OK.

PR #618, opened 2023-04-14
Fix #617: Remove springdoc-openapi-ui dependency from model

### LINKED ISSUE #617: Remove springdoc-openapi-ui dependency from model

banterCZ:


romanstrobl (approved):
Looks OK.

Issue #610, opened 2023-04-04
Upgrade Spring Boot 3


romanstrobl:
Analyze migration to Spring Boot 3 and upgrade all required project dependencies.

PR #622, opened 2023-04-17
Change compiler source to 17 and use implicit plugin versions

banterCZ:
Follow-up to #610

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -68,10 +68,6 @@
     </modules>
 
     <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <maven.compiler.source>1.8</maven.compiler.source>
-        <maven.compiler.target>1.8</maven.compiler.target>
-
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -86,9 +82,6 @@
         <bc.version>1.72</bc.version>
         <swagger-annotations-jakarta.version>2.2.8</swagger-annotations-jakarta.version>
 
-        <maven-deploy-plugin.version>3.0.0</maven-deploy-plugin.version>
-        <maven-enforcer-plugin.version>3.2.1</maven-enforcer-plugin.version>
-        <maven-jar-plugin.version>3.3.0</maven-jar-plugin.version>
         <maven-javadoc-plugin.version>3.5.0</maven-javadoc-plugin.version>
```
banterCZ, line 85: The parent uses `javadoc-plugin` version **3.4.1**, maybe we can downgrade and let it be on Spring Boot.

romanstrobl (approved):
Looks good.

PR #626, opened 2023-04-18
Fix #625: Springdoc OpenAPI UI not generated after migration to Spring Boot 3

### LINKED ISSUE #625: Springdoc OpenAPI UI not generated after migration to Spring Boot 3
We should update the dependencies.

romanstrobl:


banterCZ (approved):
Looks good.

PR #631, opened 2023-04-24
Add missing push_campaign.timestamp_sent column to liquibase

banterCZ:
Follow-up to #595


romanstrobl (approved):
Alright, I hope there is no problem with the checksums.

PR #633, opened 2023-04-24
Fix #632: Clean-up maven dependencies

### LINKED ISSUE #632: Clean-up maven dependencies
Investigate not used dependencies via `mvn dependency:analyze -DignoreNonCompile`.

banterCZ:


powerauth-push-client/pom.xml
```diff
@@ -21,32 +21,13 @@
             <artifactId>powerauth-push-model</artifactId>
         </dependency>
 
-        <!-- Spring Dependencies -->
+        <!-- Wultra dependencies -->
         <dependency>
-            <groupId>org.springframework.boot</groupId>
```
banterCZ, line 26: It does not make sense to force clients to use Spring Boot. It compiles without these dependencies.

powerauth-push-model/pom.xml
```diff
@@ -18,23 +18,16 @@
         <dependency>
             <groupId>io.getlime.core</groupId>
             <artifactId>rest-model-base</artifactId>
-            <version>${rest-base.version}</version>
-        </dependency>
-        <dependency>
-            <groupId>io.getlime.core</groupId>
-            <artifactId>rest-client-base</artifactId>
```
banterCZ, line 25: Model should not need `rest-client` dependency.

romanstrobl (approved):
Nice cleanup!

PR #641, opened 2023-07-07
Fix #640: Improve JavaDoc

### LINKED ISSUE #640: Improve JavaDoc

petrdvorak:


powerauth-push-server/src/main/java/io/getlime/push/controller/rest/InboxController.java
```diff
@@ -52,17 +52,39 @@ public class InboxController {
 
     private final InboxService inboxService;
 
+    /**
+     * Default constructor of the controller with inbox service injection.
```
banterCZ, line 56: Nitpick: this is not _default_ but _all arg_ constructor. You may skip explicit constructor and Javadoc by a lombok annotation `RequiredArgsConstructor`.

banterCZ (approved):
Looks good.

PR #673, opened 2023-09-14
Fix #672: Liquibase updates

### LINKED ISSUE #672: Liquibase updates
### Table push_app_credentials
heavily modified in 1.3.0 - leave most of the discrepancies as is, it is not possible to migrate straight from 1.30 to 1.5.0

### Table push_inbox
- inbox_id should be nullable
- the original primary key name is `push_inbox_pk` - we can allow this deviation as potential problems are theoretical and we should level notation as rest of the primary key in push domain and as per our guideline.

### Table push_inbox_app
- the original primary key name is `push_inbox_app_pk` - accepted new name

banterCZ:


romanstrobl (approved):
Looks OK.

zcgandcomp (approved):
Looks OK

PR #680, opened 2023-10-19
Fix #679: Executable war

### LINKED ISSUE #679: Executable war
Add embedded tomcat as the provided dependency (to be part of `WEB-INF/lib-provided` but not `lib`) and make it possible to run it via `java -jar`.

jandusil:


powerauth-push-server/pom.xml
```diff
@@ -175,11 +182,6 @@
             <version>${bc.version}</version>
             <scope>test</scope>
         </dependency>
-        <dependency>
-            <groupId>org.springframework.boot</groupId>
-            <artifactId>spring-boot-starter-tomcat</artifactId>
-            <scope>test</scope>
```
banterCZ, line 181: One more thing. In the parent pom, we have excluded `tomcat-embed-el` and even `spring-boot-starter-tomcat`.

banterCZ (commented):
Let's finish cleaning, please.

PR #686, opened 2023-11-07
Fix #685: HHH90000025: PostgreSQLDialect does not need to be specified explicitly

### LINKED ISSUE #685: HHH90000025: PostgreSQLDialect does not need to be specified explicitly
Fix following warning:

```
2023-11-02T10:57:09.340+01:00  WARN 56769 --- [           main] org.hibernate.orm.deprecation            : HHH90000025: PostgreSQLDialect does not need to be specified explicitly using 'hibernate.dialect' (remove the property setting and it will be selected by default)
```

jnpsk:


banterCZ (approved):
Looks good.

PR #688, opened 2023-11-23
Fix #675: Null values in extras cause a problem to APNS

### LINKED ISSUE #675: Null values in extras cause a problem to APNS
If extras map contains null values, APNS does not send the payload. It is s known Apple issue https://developer.apple.com/forums/thread/727395

banterCZ:


romanstrobl (approved):
Looks OK.

PR #706, opened 2023-12-07
Fix #705: Coverity: Eq: Problems with implementation of equals()

### LINKED ISSUE #705: Coverity: Eq: Problems with implementation of equals()
410577 [Eq: Problems with implementation of equals()](https://scan4.scan.coverity.com/doc/en/cov_checker_ref.html#static_checker_FB)
This class extends a class that defines an equals method and adds fields, but doesn't define an equals method itself. ([From SpotBugs description](https://scan4.scan.coverity.com/doc/en/spotbugs/sb_checker_ref.html#FB.EQ_DOESNT_OVERRIDE_EQUALS)) ([CWE-595](http://cwe.mitre.org/data/definitions/595.html))

banterCZ:


romanstrobl (approved):
Looks OK.

PR #741, opened 2024-01-15
Fix #681: Make create device service resilient to paralel device creation with same activationId

### LINKED ISSUE #681: Make create device service resilient to paralel device creation with same activationId
Steps to reproduce the issue:
Call push/device/create with the same `activationId`, that doesn't exist in the push server yet, multiple times in parallel.

Expected result:
A new record is created and updated several times - depending on the number of parallel calls.

Actual result:
One call creates a new record, others fail on unique constraint violation.

jnpsk:
Fix #681 by implementing retry mechanism.

- created separated method `createOrUpdateDevice()`
- the method is called repeatedly in case of the `DataIntegrityViolationException` (3 times tops by default, then the original exception is thrown)
- added missing constraint on the entity level (db schema already uses those constraint since v0.23). Necessary for the parallel creation test case.

Tested also on a running app via REST API using `seq 1 10 | xargs -n1 -P5 curl ...`

### DIFF
--- a/powerauth-push-server/src/main/java/io/getlime/push/controller/rest/PushDeviceController.java
+++ b/powerauth-push-server/src/main/java/io/getlime/push/controller/rest/PushDeviceController.java
@@ -38,6 +38,8 @@ import io.swagger.v3.oas.annotations.Operation;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.dao.DataIntegrityViolationException;
+import org.springframework.retry.support.RetryTemplate;
 import org.springframework.web.bind.annotation.*;
 
 import java.util.*;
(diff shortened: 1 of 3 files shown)


romanstrobl (approved):
I think this is a good solution, so I approve the pull request. 

However, there are few issue to consider, it can be separate improvements, too:
- I suggest to add logging so it is clear when the retries happen.
- What is the maximum number of retries? Are we relying on the default? If yes, let's document it, or otherwise set the default programatically using retry policy, so it is clear how many retries can be used.
- Have you considered using a backoff policy? In case the database is slow we may use all retry attempts when there is no backoff. So that is worth considering.

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/PushDeviceController.java
```diff
@@ -93,17 +95,32 @@ public Response createDevice(@RequestBody ObjectRequest<CreateDeviceRequest> req
...
+        RetryTemplate.builder()
+                .retryOn(DataIntegrityViolationException.class)
+                .build()
+                .execute(ctx -> createOrUpdateDevice(requestObject, appCredentials));
+
+        logger.info("The createDevice request succeeded, app ID: {}, activation ID: {}, platform: {}", requestObject.getAppId(), requestObject.getActivationId(), requestObject.getPlatform());
+        return new Response();
+    }
+
+    private Void createOrUpdateDevice(final CreateDeviceRequest requestObject, final AppCredentialsEntity appCredentials) throws PushServerException {
```
romanstrobl, line 116: Are you sure this method should return `Void` and not `void`?

romanstrobl (changes requested):
The `Void` return type seems invalid for method `createOrUpdateDevice`, it should probably be `void`.

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/PushDeviceController.java
banterCZ: Indeed, `Void` is correct, since `public final <T, E extends Throwable> T execute(RetryCallback<T, E> retryCallback) throws E {`

romanstrobl: Ah, ok, my bad.

romanstrobl (approved):

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/PushDeviceController.java
```diff
@@ -93,17 +95,32 @@ public Response createDevice(@RequestBody ObjectRequest<CreateDeviceRequest> req
...
-        String activationId = requestObject.getActivationId();
-        List<PushDeviceRegistrationEntity> devices = lookupDeviceRegistrations(appId, activationId, pushToken);
-        final AppCredentialsEntity appCredentials = findAppCredentials(appId);
-        PushDeviceRegistrationEntity device;
+
+        final AppCredentialsEntity appCredentials = findAppCredentials(requestObject.getAppId());
+
+        // In case of parallel requests to create new device, the createOrUpdateDevice may fail because of unique
+        // constraint violation. Try to repeat the save.
+        RetryTemplate.builder()
```
banterCZ, line 107: Have you considered using`@Retryable` and moving it into the service layer instead of the controller?

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/PushDeviceController.java
```diff
@@ -93,17 +95,32 @@ public Response createDevice(@RequestBody ObjectRequest<CreateDeviceRequest> req
...
+                .retryOn(DataIntegrityViolationException.class)
+                .build()
```
banterCZ, line 109: The default retry count looks good, but I agree with @romanstrobl , let's think about _backOffPolicy_.

```java
private volatile BackOffPolicy backOffPolicy = new NoBackOffPolicy();
private volatile RetryPolicy retryPolicy = new SimpleRetryPolicy(3);
```

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/PushDeviceController.java
```diff
@@ -93,17 +95,32 @@ public Response createDevice(@RequestBody ObjectRequest<CreateDeviceRequest> req
...
+    private Void createOrUpdateDevice(final CreateDeviceRequest requestObject, final AppCredentialsEntity appCredentials) throws PushServerException {
+        final String appId = requestObject.getAppId();
+        final String pushToken = requestObject.getToken();
+        final String platform = requestObject.getPlatform();
+        final String activationId = requestObject.getActivationId();
+
+        final List<PushDeviceRegistrationEntity> devices = lookupDeviceRegistrations(appId, activationId, pushToken);
+        final PushDeviceRegistrationEntity device;
         if (devices.isEmpty()) {
             // The device registration is new, create a new entity.
```
banterCZ, line 125: May we convert this comment into a log statement, at least DEBUG level?

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/PushDeviceController.java
jnpsk: Will be considered in the separate issue https://github.com/wultra/powerauth-push-server/issues/742

banterCZ (approved):
Thanks for introducing backoff.

PR #749, opened 2024-01-22
Fix #748: Switch H2 test db from file into mem

### LINKED ISSUE #748: Switch H2 test db from file into mem
It is not necessary to have a testing H2 database in a file. Moreover, it causes data incompatibility problems, when the dependency is updated.

banterCZ:


jnpsk (approved):
looks ok

PR #754, opened 2024-01-25
Fix #753: Change platform into JPA enum

### LINKED ISSUE #753: Change platform into JPA enum
Change field `PushDeviceRegistrationEntity#platform` type of `String` into `Enum` which improves type safety. Mind that using `@Enumerated(EnumType.STRING)` is not powerful enough, lower-cases need to be kept because of backward compatibility.

banterCZ:


powerauth-push-server/src/main/java/io/getlime/push/repository/model/PlatformConverter.java
```diff
@@ -0,0 +1,47 @@
...
+            case ANDROID -> "android";
+        };
+    }
+
+    @Override
+    public Platform convertToEntityAttribute(final String dbData) {
+        return switch (dbData) {
+            case "ios" -> Platform.IOS;
+            case "android" -> Platform.ANDROID;
+            default -> throw new MappingException("No mapping for platform: " + dbData);
```
romanstrobl, line 44: How is this exception being handled? Is it correct to use a Hibernate exception for a class which is not database related?

romanstrobl (approved):
This is a nice refactoring, however, let's review the error handling.

powerauth-push-server/src/main/java/io/getlime/push/repository/model/PlatformConverter.java
banterCZ: Good question. The class `package private`, is placed in the package `io.getlime.push.repository.model`, so I consider `repository` as a database package, and used by the entity `PushDeviceRegistrationEntity`.

banterCZ: At the end, I have removed the Hibernate exception.

jandusil (approved):
Good addition

PR #761, opened 2024-01-26
Fix #760: Use Lombok getter and setter annotations for entity classes

### LINKED ISSUE #760: Use Lombok getter and setter annotations for entity classes

banterCZ:


romanstrobl (approved):
Nice cleanup!

PR #763, opened 2024-01-26
Fix #762: PushMessageStatusConverter should not be a spring bean

### LINKED ISSUE #762: PushMessageStatusConverter should not be a spring bean

banterCZ:


romanstrobl (approved):
Looks OK.

PR #766, opened 2024-02-02
Fix #742: Separate service layer from PushDeviceController

### LINKED ISSUE #742: Separate service layer from PushDeviceController
Controller `PushDeviceController` contains logic that should be done in a dedicated service layer.

Consider using `@Retryable` annotation instead of `RetryTemplate` during device creation.

jnpsk:
Fix #742 by extracting validation and business logic from the `PushDeviceController`.  
There are two new service classes. `PushDeviceService` is called right away by controller, it handles validation or fetching credential entity if needed. This service class uses `DeviceRegistrationService` which finally communicates with the devices repository. The intention of this split is to reduce the complexity and to benefit from `@Retryable` annotation.

Except for the `DeviceRegistrationService#updateStatus` no significant changes were applied (just moved).  The `updateStatus` were modified to increase readability of the code. All the refactored code is now covered with tests.

The default value for retry attempts was decreased from 3 to 2, as that number is sufficient by default.

There was also a bug regarding `/push/device/create/multi` endpoint - when an activationId was repeated in single request, the server returned a constraint violation error. This is now fixed by distinct filter in `DeviceRegistrationService#createOrUpdateDevices`.


powerauth-push-server/src/main/java/io/getlime/push/service/DeviceRegistrationService.java
```diff
@@ -0,0 +1,262 @@
...
+@Slf4j
+public class DeviceRegistrationService {
+
+    private final PushDeviceRepository pushDeviceRepository;
+    private final PowerAuthClient powerAuthClient;
+
+    @Retryable(retryFor = DataIntegrityViolationException.class,
+            maxAttemptsExpression = "${powerauth.push.service.registration.retry.maxAttempts:2}",
+            backoff = @Backoff(delayExpression = "${powerauth.push.service.registration.retry.backoff:100}"))
+    public void createOrUpdateDevice(final CreateDeviceRequest requestObject, final AppCredentialsEntity appCredentials) throws PushServerException {
```
banterCZ, line 61: Should we start the transaction here? I realize that this glitch used to be already there.

powerauth-push-server/src/main/java/io/getlime/push/service/DeviceRegistrationService.java
```diff
@@ -0,0 +1,262 @@
...
+    }
+
+    public void createOrUpdateDevices(final CreateDeviceForActivationsRequest request, final AppCredentialsEntity appCredentials) throws PushServerException {
+        final String appId = request.getAppId();
+        final String pushToken = request.getToken();
+        final MobilePlatform platform = request.getPlatform();
+        final List<String> activationIds = request.getActivationIds();
+
+        // Initialize loop variables.
+        AtomicBoolean registrationFailed = new AtomicBoolean(false);
```
banterCZ, line 101: May it be `final`?

powerauth-push-server/src/main/java/io/getlime/push/service/DeviceRegistrationService.java
```diff
@@ -0,0 +1,262 @@
...
+     * </ul>
+     * @param appId Application ID.
+     * @param activationId Activation ID.
+     * @param pushToken Push token.
+     * @return List of found device registration entities.
+     */
+    private List<PushDeviceRegistrationEntity> lookupDeviceRegistrations(String appId, String activationId, String pushToken) throws PushServerException {
+        List<PushDeviceRegistrationEntity> deviceRegistrations;
+        // At first, lookup the device registrations by match on activationId and pushToken.
+        deviceRegistrations = pushDeviceRepository.findByActivationIdAndPushToken(activationId, pushToken);
```
banterCZ, line 179: It could be inlined with the declaration. Moreover rewriting to `final` would be less error-prone.

powerauth-push-server/src/test/java/io/getlime/push/service/PushDeviceServiceTest.java
```diff
@@ -0,0 +1,199 @@
...
+        assertEquals("OK", response.getStatus());
+    }
+
+    @Test
+    void testDeleteDevice_invalidRequest() {
+        final DeleteDeviceRequest request = new DeleteDeviceRequest();
+        request.setAppId("app1");
+
+        final PushServerException exception = assertThrows(PushServerException.class,
+                () ->tested.deleteDevice(request));
```
banterCZ, line 195:
```suggestion
                () -> tested.deleteDevice(request));
```

banterCZ (approved):
Thanks for moving the logic out of the controller.

PR #779, opened 2024-02-19
Fix #772: Notification Settings FAQ

### LINKED ISSUE #772: Notification Settings FAQ
Our support is often contacted with issues about push notification misconfiguration. Prepare a FAQ document explaining the typical problems and how to solve them.

banterCZ:


docs/FAQ.md
```diff
@@ -0,0 +1,36 @@
+# Frequently Asked Questions (FAQ)
+
+
+## APNs
+
+When iOS notifications do not work, double-check APNS configuration.
+Mainly the host, the key, and the bundle.
+
+APNs expose two hosts, production `api.push.apple.com` and development `api.sandbox.push.apple.com`.
```
zcgandcomp, line 9: This is not configurable by the user, so do not disclose it can lead to more misunderstanding

docs/FAQ.md
```diff
@@ -0,0 +1,36 @@
...
+
+Push Server uses [property powerauth.push.service.apns.useDevelopment](./Configuration-Properties#apns-configuration) configuring the host globally.
+Value `false` represents the production environment.
+
+
+## Application Configuration
+
+The host may be overridden per application in the database table [push_app_credentials.ios_environment](./Push-Server-Database#push-service-credentials-table).
+
+You may use the following values in the column `ios_environment`:
```
zcgandcomp, line 24: this is a reducancy to the API configuration - and we do not want to enable DB mingling.

docs/FAQ.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+## APNs
+
+When iOS notifications do not work, double-check APNS configuration.
+Mainly the host, the key, and the bundle.
+
+APNs expose two hosts, production `api.push.apple.com` and development `api.sandbox.push.apple.com`.
+
+The typical misconfiguration is the usage of the dev key or the dev bundle with the production host or vice versa.
```
zcgandcomp, line 11: Like the idea is good, but it should be done differently.

There should be errors from the connection to APNS and required action on our side - this is to generic.

zcgandcomp (changes requested):
Sorry good idea, we just need to do it more specific

docs/FAQ.md
banterCZ: @kober32 @petrdvorak Are we really able to distinguish this kind of error?

zcgandcomp: https://developer.apple.com/documentation/usernotifications/setting_up_a_remote_notification_server/handling_notification_responses_from_apns

Most typical cases:

BadDeviceToken - problem with the environment or with a token.

DeviceTokenNotForTopic - problem with bundle Id or topic (they do not match)

banterCZ: Any idea how to rephrase it? My point was, that quite often, customers use the dev key against the production host.

banterCZ:
Let's close the PR, since it is not approved.

PR #804, opened 2024-03-01
Fix #803: String#replace should be preferred to String#replaceAll

### LINKED ISSUE #803: String#replace should be preferred to String#replaceAll
See `CaCertUtil`.

https://rules.sonarsource.com/java/RSPEC-5361/

banterCZ:


romanstrobl (approved):
Agreed.

PR #808, opened 2024-03-14
Fix #807: Fix error handling for FCM errors

### LINKED ISSUE #807: Fix error handling for FCM errors
In responses to FCM API calls, we can see:

```json
{
  "error": {
    "code": 404,
    "message": "Requested entity was not found.",
    "status": "NOT_FOUND",
    "details": [
      {
        "@type": "type.googleapis.com/google.firebase.fcm.v1.FcmError",
        "errorCode": "UNREGISTERED"
      }
    ]
  }
}
```

```json
{
  "error": {
    "code": 400,
    "message": "The registration token is not a valid FCM registration token",
    "status": "INVALID_ARGUMENT",
    "details": [
      {
        "@type": "type.googleapis.com/google.firebase.fcm.v1.FcmError",
        "errorCode": "INVALID_ARGUMENT"
      }
    ]
  }
}
```

Both are legit states, see:
https://firebase.google.com/docs/cloud-messaging/manage-tokens#detect-invalid-token-responses-from-the-fcm-backend

As a result, these messages should be logged just as `INFO`, not `WARN`.

Also, this error handling does not seem correct:
[https://github.com/wultra/powerauth-push-server/blob/cbe9f710dfe009c1151c6687abf09[…]er/src/main/java/io/getlime/push/service/PushSendingWorker.java](https://github.com/wultra/powerauth-push-server/blob/cbe9f710dfe009c1151c6687abf0988b15e4c379/powerauth-push-server/src/main/java/io/getlime/push/service/PushSendingWorker.java#L186)

In case of a correctly formed request, we should interpret `INVALID_ARGUMENT` as an invalid token and remove it. The request is likely correctly formatted as we use Firebase Admin by Google. We currently keep such token stored...

jandusil:


banterCZ (commented):
May we have a test for that?

powerauth-push-server/src/main/java/io/getlime/push/service/PushSendingWorker.java
```diff
@@ -200,7 +200,7 @@ void sendMessageToAndroid(final FcmClient fcmClient, final PushMessageBody pushM
                         callback.didFinishSendingMessage(PushSendingCallback.Result.PENDING);
                         return;
                     }
-                    case SENDER_ID_MISMATCH, THIRD_PARTY_AUTH_ERROR, INVALID_ARGUMENT -> {
+                    case SENDER_ID_MISMATCH, THIRD_PARTY_AUTH_ERROR-> {
```
banterCZ, line 203:
```suggestion
                    case SENDER_ID_MISMATCH, THIRD_PARTY_AUTH_ERROR -> {
```

banterCZ (approved):
Thanks for the test.

PR #825, opened 2024-04-22
Fix #824: Refactor BouncyCastle dependency

### LINKED ISSUE #824: Refactor BouncyCastle dependency
Use transient dependency from `powerauth-crypto`. Depends on https://github.com/wultra/powerauth-crypto/issues/605

banterCZ:
Use transient dependency from `powerauth-crypto` which is scope `test`.

- [x] Depends on https://github.com/wultra/powerauth-crypto/pull/606


jnpsk (approved):
looks ok

PR #840, opened 2024-05-31
Fix #839: Throwable and Error should not be caught

### LINKED ISSUE #839: Throwable and Error should not be caught

jandusil:


powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -182,7 +182,7 @@ private PushSendingCallback createPushSendingCallback(final Mode mode, final Pus
                     }
                 }
                 platformResult.setTotal(platformResult.getTotal() + 1);
-            } catch (Throwable t) {
+            } catch (Exception t) {
```
banterCZ, line 185: I suggest to rename `t` into `e`.

banterCZ (approved):

PR #862, opened 2024-07-08
Fix #861: SQL error after DB column escaping

### LINKED ISSUE #861: SQL error after DB column escaping
The column escaping introduced in https://github.com/wultra/powerauth-push-server/pull/859/files does not work properly on Oracle due to enforcing case sensitivity.

We should probably go back to the original solution which was renaming columns, which has no side effects, it only slightly complicates the migration to new release.

Log:
```
app-1      | 2024-07-03T13:08:54.639Z ERROR 218 --- [powerauth-java-server] [io-8080-exec-14] [0395b0626facc7544e7e4fc1383731fe-199b8d50fd8766c9] o.h.engine.jdbc.spi.SqlExceptionHelper   : ORA-00904: "type": invalid identifier
app-1      |
app-1      | https://docs.oracle.com/error-help/db/ora-00904/
app-1      | 2024-07-03T13:08:54.646Z ERROR 218 --- [powerauth-java-server] [io-8080-exec-14] [0395b0626facc7544e7e4fc1383731fe-199b8d50fd8766c9] i.g.p.e.DefaultExceptionHandler          : could not execute statement [ORA-00904: "type": invalid identifier
app-1      |
app-1      | https://docs.oracle.com/error-help/db/ora-00904/] [insert into push_inbox (body,inbox_id,"read",subject,summary,timestamp_created,timestamp_read,"type",user_id,id) values (?,?,?,?,?,?,?,?,?,?)]; SQL [insert into push_inbox (body,inbox_id,"read",subject,summary,timestamp_created,timestamp_read,"type",user_id,id) values (?,?,?,?,?,?,?,?,?,?)]
app-1      |
app-1      | org.springframework.dao.InvalidDataAccessResourceUsageException: could not execute statement [ORA-00904: "type": invalid identifier
app-1      |
```

romanstrobl:
Due to problems with column name escaping in Oracle caused by enforced case sensitivity I had to rename the two columns which are problematic keywords in MSSQL.

### DIFF
--- a/docs-private/Developer-How-To-Start.md
+++ b/docs-private/Developer-How-To-Start.md
@@ -21,3 +21,22 @@ Others (like URL, username, password) depend on your environment.
 ```shell
 liquibase --changelog-file=./docs/db/changelog/changesets/powerauth-push-server/db.changelog-module.xml --url=jdbc:postgresql://localhost:5432/powerauth --username=powerauth status
 ``` 
+
+#### PostgreSQL
+
+```shell
+liquibase --changeLogFile=./docs/db/changelog/changesets/powerauth-push-server/db.changelog-module.xml --output-file=./docs/sql/postgresql/migration_1.7.0_1.8.0.sql updateSQL --url=offline:postgresql
+```
+
+#### Oracle
+
+```shell
+liquibase --changeLogFile=./docs/db/changelog/changesets/powerauth-push-server/db.changelog-module.xml --output-file=./docs/sql/oracle/migration_1.7.0_1.8.0.sql updateSQL --url=offline:oracle
+```
+
+
+#### MSSQL
+
+```shell
+liquibase --changeLogFile=./docs/db/changelog/changesets/powerauth-push-server/db.changelog-module.xml --output-file=./docs/sql/mssql/migration_1.7.0_1.8.0.sql updateSQL --url=offline:mssql
+```
--- a/docs-private/Developer-How-To-Start.md
+++ b/docs-private/Developer-How-To-Start.md
@@ -22,6 +22,8 @@ Others (like URL, username, password) depend on your environment.
 liquibase --changelog-file=./docs/db/changelog/changesets/powerauth-push-server/db.changelog-module.xml --url=jdbc:postgresql://localhost:5432/powerauth --username=powerauth status
 ``` 
 
+To generate SQL script run following command:
+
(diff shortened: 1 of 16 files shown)


docs-private/Developer-How-To-Start.md
```diff
@@ -21,3 +21,22 @@ Others (like URL, username, password) depend on your environment.
 ```shell
 liquibase --changelog-file=./docs/db/changelog/changesets/powerauth-push-server/db.changelog-module.xml --url=jdbc:postgresql://localhost:5432/powerauth --username=powerauth status
 ``` 
+
```
zcgandcomp, line 24: Maven has liquibase profile that process the liquibase for development.  We can reference the profile (as the profile is standard in our components)
Or we provide the generated SQL by ourselves.

I suggest not to complicate and generate SQL itself - this should be discouraged.

docs/PowerAuth-Push-Server-1.8.0.md
```diff
@@ -0,0 +1,16 @@
+# Migration from 1.7.x to 1.8.x
+
+This guide contains instructions for migration from PowerAuth Push Server version `1.7.x` to version `1.8.x`.
+
+
+## Database Changes
+
+For convenience, you can use liquibase for your database migration.
+
+**Important: Upgrading to version 1.8.x includes column renaming, resulting in incompatible changes. Therefore, ensure all Push Server nodes are upgraded simultaneously during a scheduled service window.**
```
zcgandcomp, line 10: I suggest stating that this is for the Inbox feature. For proper impact analysis.

zcgandcomp (changes requested):
Remarks for documentation the implementation looks OK.

docs-private/Developer-How-To-Start.md
romanstrobl: This is just developer documentation in `docs-private` folder. We have it in other components, too. If you want I can remove it.

docs/PowerAuth-Push-Server-1.8.0.md
romanstrobl: Good idea.

jnpsk (approved):
Looks ok. I just noticed the reference for the migration instruction is missing in the `Migration-Instructions.md `file.

zcgandcomp (approved):
Looks OK.

PR #880, opened 2024-10-01
Fix #841: Add warning when push notification cannot be sent

### LINKED ISSUE #841: Add warning when push notification cannot be sent due to an invalid userId or userId+activationId combination
Currently if `userId` or combination of `userId` and `activationId` doesn't have any activation to send push notification to, we do not send the push, but push server doesn't provide any warning.

Add to log WARN message that no device was found to send the push notification to.

(Alternatively we can raise an exception if the combination is not valid - TBD)

banterCZ:
Add warning when push notification cannot be sent due to an invalid userId or userId+activationId combination

### DIFF
--- a/powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
+++ b/powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
@@ -261,16 +261,24 @@ public class PushMessageSenderService {
     }
 
     // Return list of devices related to given user or activation ID (if present). List of devices is related to particular application as well.
-    private List<PushDeviceRegistrationEntity> getPushDevices(Long id, String userId, String activationId) throws PushServerException {
+    private List<PushDeviceRegistrationEntity> getPushDevices(Long appCredentialsId, String userId, String activationId) throws PushServerException {
         if (userId == null || userId.isEmpty()) {
             logger.error("No userId was specified");
             throw new PushServerException("No userId was specified");
         }
+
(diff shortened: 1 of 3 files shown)


powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -261,16 +261,24 @@ private AppCredentialsEntity getAppCredentials(String appId) throws PushServerEx
...
         if (activationId != null) { // in case the message should go to the specific device
-            return pushDeviceRepository.findByUserIdAndAppCredentialsIdAndActivationId(userId, id, activationId);
+            devices = pushDeviceRepository.findByUserIdAndAppCredentialsIdAndActivationId(userId, appCredentialsId, activationId);
         } else {
-            return pushDeviceRepository.findByUserIdAndAppCredentialsId(userId, id);
+            devices = pushDeviceRepository.findByUserIdAndAppCredentialsId(userId, appCredentialsId);
+        }
+
+        if (devices.isEmpty()) {
+            throw new PushServerException("No device found for userId=%s, appCredentialsId=%s, activationId=%s".formatted(userId, appCredentialsId, activationId));
```
zcgandcomp, line 278: For now please just log the warning. Thank you

zcgandcomp (changes requested):

zcgandcomp (approved):
ok, nice. Let's followup if the exception is better later on.

PR #885, opened 2024-10-10
Fix #884: clearing the cache of the push client works only on one node

### LINKED ISSUE #884: caching of push client is not sensitive for changes
Step to reproduce:
Create FCM config for an application.
Send push notifications for the registration of the application for the Android platform. (client is stored in cache)
Create APNS configuration for the application.
Send push notifications for the registration of the application for the iOS platform. (client is read from cache)

Expected result: 
The push notification is sent

Actual result:
No push is sent. An error in the log:
`Push message cannot be sent to APNS because APNS is not configured in push server`

banterCZ:
- limit expiration time


powerauth-push-server/src/main/java/io/getlime/push/configuration/StorageConfiguration.java
```diff
@@ -16,25 +16,32 @@
...
 /**
- * Configuration class for in memory maps
+ * Configuration class for storage.
  *
  * @author Martin Tupy, martin.tupy.work@gmail.com
  */
 @Configuration
-public class StorageMapConfiguration {
+@Slf4j
+public class StorageConfiguration {
```
banterCZ, line 35: I would rather call it `PushServiceConfiguration` but there already is such a class, which should be named `PushServiceConfigurationProperties`. Anyway, let's postpone to another PR.

powerauth-push-server/src/main/java/io/getlime/push/service/batch/storage/ItemStorage.java
```diff
@@ -37,18 +37,6 @@ public interface ItemStorageMap<K,V> {
      */
     void put(K key, V value);
 
-    /**
```
banterCZ, line 40: Not used at all.

powerauth-push-server/src/main/java/io/getlime/push/service/batch/storage/ItemStorage.java
```diff
@@ -21,7 +21,7 @@
  * @param <K> Key class for item storage.
  * @param <V> Value class for item storage.
  */
-public interface ItemStorageMap<K,V> {
+public interface ItemStorage<K,V> {
```
banterCZ, line 24: The interface should not say much about the underlying implementation. Anyway, we should switch to `Cache` later on instead of the custom one, to enable such a feature as `#get(K key, Function<K, V> mappingFunction)`.

powerauth-push-server/src/main/resources/application.properties
```diff
@@ -53,6 +53,8 @@ powerauth.push.service.registration.multipleActivations.enabled=false
 powerauth.push.service.registration.retry.backoff=100
 powerauth.push.service.registration.retry.maxAttempts=2
 
+powerauth.push.service.clients.cache.expireAfterWrite=24h
```
banterCZ, line 56: @zcgandcomp What is the reasonable default?

powerauth-push-server/src/main/java/io/getlime/push/service/batch/storage/CampaignMessageStorage.java
```diff
@@ -28,7 +28,7 @@
  *
  * @author Petr Dvorak, petr@wultra.com
  */
-public class CampaignMessageStorageMap implements ItemStorageMap<Long, PushCampaignEntity> {
+public class CampaignMessageStorage implements ItemStorage<Long, PushCampaignEntity> {
```
banterCZ, line 31: 🤔 Should we set expiration even for campaigns?

banterCZ: Ok, let's ignore expiration here because the campaign is cached for a single run only.

powerauth-push-server/src/main/resources/application.properties
banterCZ: It seems that 24 hours is too long, but 5 minutes is too short because initializing the client is quite heavy. Let's consider some smart reloadable cache.

docs/Push-Server-Database.md
```diff
@@ -92,25 +92,29 @@ CREATE TABLE push_app_credentials (
     ios_bundle VARCHAR(255),
     ios_environment VARCHAR(32),
     android_private_key BYTEA,
-    android_project_id VARCHAR(255)
+    android_project_id VARCHAR(255),
```
banterCZ, line 95: Let's remove SQL completely in a different PR. This update just makes it consistent.

zcgandcomp (changes requested):
few issues

powerauth-push-server/src/main/java/io/getlime/push/service/AppRelatedPushClientCacheLoader.java
```diff
@@ -0,0 +1,104 @@
...
+
+        final LocalDateTime lastUpdatedInDb = credentials.getTimestampLastUpdated();
+        final LocalDateTime lastUpdatedInCache = oldValue.getAppCredentials().getTimestampLastUpdated();
+        if (Objects.equals(lastUpdatedInCache, lastUpdatedInDb)) {
+            logger.debug("LastUpdated is same for app: {}", key);
+            return oldValue;
+        }
+
+        logger.debug("LastUpdated differs for app: {}", key);
+        return load(key);
```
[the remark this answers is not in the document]
banterCZ: A good point. Normally, it should be cached by the entity manager, but since it is loaded async, it will load the record twice.

jnpsk (approved):
looks good, nice solution

powerauth-push-server/src/main/java/io/getlime/push/repository/model/AppCredentialsEntity.java
```diff
@@ -112,4 +113,10 @@ public class AppCredentialsEntity implements Serializable {
     @Column(name = "hms_client_secret")
     private String hmsClientSecret;
 
+    @Column(name = "timestamp_created", nullable = false, columnDefinition = "TIMESTAMP DEFAULT CURRENT_TIMESTAMP")
```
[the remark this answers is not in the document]
banterCZ: This is a hint for h2-generated db in tests.

powerauth-push-server/src/main/java/io/getlime/push/service/AppRelatedPushClientCacheLoader.java
```diff
@@ -65,22 +65,23 @@ public AppRelatedPushClient reload(final String appId, final AppRelatedPushClien
...
     }
 
     @Override
-    public AppRelatedPushClient load(final String key) throws Exception {
-        final AppCredentialsEntity credentials = appCredentialsRepository.findFirstByAppId(key).orElse(null);
+    public AppRelatedPushClient load(final String appId) throws Exception {
+        final AppCredentialsEntity credentials = appCredentialsRepository.findFirstByAppId(appId).orElse(null);
         if (credentials == null) {
-            logger.debug("AppCredentials does not exist for app: {}", key);
+            logger.debug("AppCredentials does not exist for app: {}", appId);
```
zcgandcomp, line 75: I suggest raising it as a warning, as this might not be a breaking problem, but very probably it can be. Same for reload.

zcgandcomp (changes requested):
Small change for logging and we are clear to go.

zcgandcomp (approved):
Looks OK

PR #900, opened 2024-10-15
Fix #899: Remove SQL schema creation from markdown

### LINKED ISSUE #899: Remove SQL schema creation from markdown
Remove schema creation from markdown and link SQL files only to avoid duplication.

banterCZ:


romanstrobl (approved):
Looks OK.

PR #905, opened 2024-10-30
Fix #904: Add new platform enum values and deprecate old ones

### LINKED ISSUE #904: Add new platform enum values and deprecate old ones
Add `apns`, `fcm` and `hms` to the platform enums, store new configuration in new columns, and make the change in a compatible way. Migrate existing configuration where necessary. Update REST API and services to handle new enums.

romanstrobl:


docs/Push-Server-Database.md
```diff
@@ -35,7 +35,7 @@ Stores push tokens specific for a given device.
 | `activation_id`             | VARCHAR(37)  | index                             | Activation ID associated with given push token record.                                                                                |
 | `user_id`                   | INTEGER      | index                             | Associated user ID.                                                                                                                   |
 | `app_id`                    | INTEGER      | index                             | Associated application ID.                                                                                                            |
-| `platform`                  | VARCHAR(30)  | -                                 | Mobile OS Platform ("ios", "android").                                                                                                |
+| `platform`                  | VARCHAR(30)  | -                                 | Mobile OS Platform ("apns", "fcm", "hms").                                                                                             |
```
zcgandcomp: Strictly speaking we still supporting the `ios` and `android` values - for old devices

docs/db/changelog/changesets/powerauth-push-server/1.10.x/20241029-add-new-platforms.xml
```diff
@@ -0,0 +1,118 @@
...
+        </preConditions>
+        <comment>Migrate existing android_* columns to fcm_* columns</comment>
+        <update tableName="push_app_credentials">
+            <column name="fcm_private_key" valueComputed="android_private_key"/>
+            <column name="fcm_project_id" valueComputed="android_project_id"/>
+        </update>
+    </changeSet>
+
+    <!-- Migrate device registrations: ios to apns -->
+    <changeSet id="5" logicalFilePath="powerauth-push-server/1.10.x/20241029-migrate-ios-to-apns.xml" author="Roman Strobl">
```
zcgandcomp: The above migration is really nice. But I would not migrate the devices itself. It will work anyway as I saw in the code, so is it worth to update the database? BTW the performance impact of the update would be quite big.

powerauth-push-model/src/main/java/io/getlime/push/model/entity/PushMessageSendResult.java
```diff
@@ -30,27 +30,27 @@
 public class PushMessageSendResult extends BasePushMessageSendResult {
```
zcgandcomp, line 30: Do we need this class at all? I failed to see the logic in using this class. Let's discuss please.

powerauth-push-model/src/main/java/io/getlime/push/model/response/GetApplicationDetailResponse.java
```diff
@@ -33,33 +33,33 @@ public class GetApplicationDetailResponse {
...
 
     /**
-     * iOS key ID.
+     * APNs key ID.
      */
-    private String iosKeyId;
+    private String apnsKeyId;
 
     /**
-     * iOS team ID.
```
zcgandcomp, line 46: we should be consistent we refer to this as ios on other places, here as apns. It's actually iOS team ID - needed for APNS setup but existing without APNS (same for bundle, key is actually apns key)

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/AdministrationController.java
```diff
@@ -165,13 +171,36 @@ public Response updateIos(@RequestBody ObjectRequest<UpdateIosRequest> request)
         return new Response();
     }
 
+    /**
+     * Update APNs configuration.
+     * @param request Update APNs configuration request.
+     * @return Response.
+     * @throws PushServerException Thrown when application credentials entity could not be found or request validation fails.
+     */
+    @RequestMapping(value = "apns/update", method = { RequestMethod.POST, RequestMethod.PUT })
```
zcgandcomp: Should we not update the API? At least the new one.

the `/admin/app/apns` is correct (and subsequent changes to the remove API.  Not necessary, but up to discussion.

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/AdministrationController.java
```diff
@@ -236,13 +333,30 @@ public Response removeAndroid(@RequestBody ObjectRequest<RemoveAndroidRequest> r
      * @throws PushServerException Thrown when application credentials entity could not be found or request validation fails.
      */
     @RequestMapping(value = "huawei/update", method = { RequestMethod.POST, RequestMethod.PUT })
+    @Operation(summary = "Update Huawei configuration (deprecated)", description = "Update Huawei configuration endpoint (deprecated), use the /admin/app/hms/update endpoint")
     public Response updateHuawei(@Valid @RequestBody ObjectRequest<UpdateHuaweiRequest> request) throws PushServerException {
         final UpdateHuaweiRequest requestObject = request.getRequestObject();
-        logger.info("Received update Huawei request, application ID: {}", requestObject.getAppId());
+        logger.info("Received updateHuawei request, application ID: {}", requestObject.getAppId());
```
zcgandcomp: We can apply logging guide i.e.
        logger.info("action: {}, state: {}, applicationId: {}", "updateHuawei", "initiated", requestObject.getAppId());
and in analogy to the whole controller

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -96,31 +96,31 @@ public BasePushMessageSendResult sendPushMessage(final String appId, final Mode
...
+                    if (platform == Platform.IOS || platform == Platform.APNS) {
                         if (pushClient.getApnsClient() == null) {
                             logger.error("Push message cannot be sent to APNS because APNS is not configured in push server.");
                             arriveAndDeregisterPhaserForMode(phaser, mode);
                             continue;
                         }
-                        final PushMessageSendResult.PlatformResult platformResult = sendResult.getIos();
-                        pushSendingWorker.sendMessageToIos(pushClient.getApnsClient(), pushMessage.getBody(), pushMessage.getAttributes(), pushMessage.getPriority(), device.getPushToken(), pushClient.getAppCredentials().getIosBundle(), createPushSendingCallback(mode, device, platformResult, pushMessageObject, phaser));
-                    } else if (platform == Platform.ANDROID) {
+                        final PushMessageSendResult.PlatformResult platformResult = sendResult.getApns();
```
zcgandcomp, line 104: Here I fail to see the purpose for the platform result.

Shouldn't we rename the sendMessageToIos to sendMessageToApns?

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -96,31 +96,31 @@ public BasePushMessageSendResult sendPushMessage(final String appId, final Mode
...
+                    } else if (platform == Platform.ANDROID || platform == Platform.FCM) {
                         if (pushClient.getFcmClient() == null) {
                             logger.error("Push message cannot be sent to FCM because FCM is not configured in push server.");
                             arriveAndDeregisterPhaserForMode(phaser, mode);
                             continue;
                         }
                         final String token = device.getPushToken();
-                        final PushMessageSendResult.PlatformResult platformResult = sendResult.getAndroid();
+                        final PushMessageSendResult.PlatformResult platformResult = sendResult.getFcm();
                         pushSendingWorker.sendMessageToAndroid(pushClient.getFcmClient(), pushMessage.getBody(), pushMessage.getAttributes(), pushMessage.getPriority(), token, createPushSendingCallback(mode, device, platformResult, pushMessageObject, phaser));
```
zcgandcomp: Same rename here?

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -202,7 +202,7 @@ public void sendCampaignMessage(final String appId, Platform platform, final Str
 
         switch (platform) {
             case IOS ->
```
zcgandcomp: We should need to add a new platform, right?

zcgandcomp (changes requested):
Mostly just issues for discussion. I would really avoid migration of device registration - it is not necessary and can bring performance issues.

I kinda think this change is without outage on multiple nodes - and good job using liquibase to migrate the credentials.

docs/PowerAuth-Push-Server-1.10.0.md
```diff
@@ -0,0 +1,52 @@
+# Migration from 1.9.x to 1.10.x
```
banterCZ, line 1: This file should be linked from `Migration-Instructions.md`.

powerauth-push-model/src/main/java/io/getlime/push/model/request/RemoveApnsRequest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Application ID.
+     */
+    @NotBlank
+    @Schema(description = "Application ID.")
+    private String appId;
+
+    /**
+     * Default constructor.
```
banterCZ: Strictly speaking, this is a no-arg constructor.
```suggestion
     * No-arg constructor.
```

powerauth-push-model/src/main/java/io/getlime/push/model/request/RemoveFcmRequest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Application ID.
+     */
+    @NotBlank
+    @Schema(description = "Application ID.")
+    private String appId;
+
+    /**
+     * Default constructor.
```
banterCZ:
```suggestion
     * No-arg constructor.
```

powerauth-push-model/src/main/java/io/getlime/push/model/request/UpdateApnsRequest.java
```diff
@@ -0,0 +1,98 @@
...
+
+    /**
+     * Base64 encoded private key.
+     */
+    @NotBlank
+    @Schema(description = "Base64 encoded private key.")
+    private String privateKeyBase64;
+
+    /**
+     * Default constructor.
```
banterCZ:
```suggestion
     * No-arg constructor.
```

powerauth-push-model/src/main/java/io/getlime/push/model/validator/RemoveApnsRequestValidator.java
```diff
@@ -0,0 +1,43 @@
...
+package io.getlime.push.model.validator;
+
+import io.getlime.push.model.request.RemoveApnsRequest;
+
+/**
+ * Validator class for remove APNs request.
+ *
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+public class RemoveApnsRequestValidator {
```
banterCZ: 🤔 Couldn't we use bean validation?

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -562,12 +566,37 @@ public Response updateIos(String appId, String bundle, String keyId, String team
...
+     * @param keyId The APNs key ID record.
+     * @param teamId The APNs team ID record.
+     * @param environment The APNs environment.
+     * @param privateKey The APNs private key bytes.
+     * @return Response from server.
+     * @throws PushServerClientException Thrown when communication with Push Server fails.
+     */
+    public Response updateApns(String appId, String bundle, String keyId, String teamId, ApnsEnvironment environment, byte[] privateKey) throws PushServerClientException {
+        final String privateKeyBase64 = Base64.getEncoder().encodeToString(privateKey);
+        final UpdateApnsRequest request = new UpdateApnsRequest(appId, bundle, keyId, teamId, environment, privateKeyBase64);
```
banterCZ: May we replace many arguments with a single request object? The objection is related to the constructor, which has too many parameters of the same type (it could be replaced with a builder).

banterCZ (commented):
Good job! I have added severals comments to consider.

docs/db/changelog/changesets/powerauth-push-server/1.10.x/20241029-add-new-platforms.xml
romanstrobl: Ok, I will remove the migration of device registrations.

powerauth-push-model/src/main/java/io/getlime/push/model/entity/PushMessageSendResult.java
romanstrobl: Yes, this class is used in the `sendPushMessage` method.

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/AdministrationController.java
romanstrobl: I am not sure I understand what you mean by this comment.

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/AdministrationController.java
romanstrobl: Ok, I will change the logging in whole class.

powerauth-push-model/src/main/java/io/getlime/push/model/validator/RemoveApnsRequestValidator.java
romanstrobl: Yes, I can switch to bean validation for all validators.

powerauth-push-server/src/main/java/io/getlime/push/controller/rest/AdministrationController.java
zcgandcomp: Apologies, it seems I accidentally replaced the text with the pasting of the example.

We can leave the REST API pattern as-is. 
But we can change to a more standard approach - at least for the new (non-deprecated API)
i.e., `POST /admin/app/apns` instead of `POST /admin/app/apns/update`
With the following changes to the remove API, i.e., `DELETE /admin/app/apns` instead of `DELETE /admin/app/apns/remove`

romanstrobl: OK, I can change it.

romanstrobl: For the `DELETE` methods, we will need to use `@RequestParam String appId` because request bodies may not be always supported for `DELETE` methods and the `POST` method is removed.

powerauth-push-model/src/main/java/io/getlime/push/model/entity/PushMessageSendResult.java
zcgandcomp: Yes, I can see the usage in the code, the purpose of the usage eludes me.

zcgandcomp (approved):
Logic is OK. I still fail to see the purpose of the class  `PushMessageSendResult`, but the usage of the class is the same as before.

romanstrobl:
The `PushMessageSendResult` contains detailed results push message delivery results, which are filled in synchronous mode both for a single push message delivery and for push message batch delivery. The response contains `BasePushMessageSendResult`, so that same response class can be used for both synchronous and asynchronous modes. The responses should be correct for batches, because `PushMessageSendResult` extends `BasePushMessageSendResult` and Jackson serialization will serialize the child class in case details are available. Jackson serializes the actual runtime type of the object.

PR #907, opened 2024-11-08
Fix #906: Allow customization of APNs environment on push device level

### LINKED ISSUE #906: Allow customization of APNs environment on push device level
Make following changes to allow the environment setting to be more granural on device level:

- Allow customization of APNs environment on push device level.
  - Mobile SDK will send the correct environment during push device registration, add an `environment` parameter.
  - Store the `environment` parameter with the push device registration in the database.
- In case the environment is not set on device level (`null` value for old registrations), use application credentials configuration in table `push_app_credentials`, column `apns_environment`.
- In case the environment is not set on application configuration level either, use the global server settings when sending the push messages configured using property `powerauth.push.service.apns.useDevelopment` as a fallback.
  - Do not allow sending messages to devices with `development` environment in case the server is configured in `production` mode (`powerauth.push.service.apns.useDevelopment=false`).
  - For this case, just log a warning. The server has to be changed to `development` mode to allow sending of such messages.
- Cache two APNs clients for separating `production` and `development` clients and use the client based on the environment.
  - Choose the right client based on device configuration (highest priority), application credentials configuration or global setting as a fallback.

romanstrobl:


powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -121,15 +121,17 @@ public boolean createDevice(String appId, String token, MobilePlatform platform)
      * @param appId PowerAuth application app ID.
      * @param token Token received from the push service provider (APNs, FCM).
      * @param platform Mobile platform (APNs, FCM, HMS).
+     * @param environment APNs environment (optional).
      * @param activationId PowerAuth activation ID.
      * @return True if device registration was successful, false otherwise.
      * @throws PushServerClientException In case of network, response / JSON processing, or other IO error.
      */
-    public boolean createDevice(String appId, String token, MobilePlatform platform, String activationId) throws PushServerClientException {
+    public boolean createDevice(String appId, String token, MobilePlatform platform, ApnsEnvironment environment, String activationId) throws PushServerClientException {
```
banterCZ: Strictly speaking, this is not binary compatible; on the other hand, it is easy for the clients to fix. Anyway, I suggest deprecating these methods and introducing a request parameter. Could be handled by a different PR.

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -194,14 +197,20 @@ public void sendCampaignMessage(String appId, Platform platform, String token, P
...
 
         final PushMessageEntity pushMessageObject = pushMessageDAO.storePushMessageObject(pushMessageBody, attributes, userId, activationId, deviceId);
 
         switch (platform) {
-            case IOS, APNS ->
-                    pushSendingWorker.sendMessageToApns(pushClient.getApnsClient(), pushMessageBody, attributes, priority, token, pushClient.getAppCredentials().getApnsBundle(), createPushSendingCallback(token, pushMessageObject, pushClient));
+            case IOS, APNS -> {
+                final ApnsEnvironment apnsEnvironment = environment != null ? resolveApnsEnvironment(environment.getKey()) : resolveApnsEnvironment(null);
+                if (apnsEnvironment == null) {
+                    logger.error("Campaign push message cannot be sent because APNs environment configuration failed. Check configuration of application property 'powerauth.push.service.apns.useDevelopment'.");
```
banterCZ: 🤔 Shouldn't we return from this branch or wrap the next lines into an `else` branch?

powerauth-push-server/src/main/resources/application.properties
```diff
@@ -56,7 +56,7 @@ powerauth.push.service.registration.retry.maxAttempts=2
 powerauth.push.service.clients.cache.refreshAfterWrite=5m
 
 # APNs Configuration
-powerauth.push.service.apns.useDevelopment=true
+powerauth.push.service.apns.useDevelopment=false
```
banterCZ: 👀 Shouldn't we document this change?

banterCZ (commented):
In general, looks good. Let's wait for documentation and tests.

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
romanstrobl: Yes, good catch.

powerauth-push-server/src/main/resources/application.properties
romanstrobl: Yes, I will prepare documentation.

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
romanstrobl: Ok, I can make the change in a compatible way and use deprecation and write a migration note.

powerauth-push-server/src/main/resources/application.properties
romanstrobl: I will revert this change, because it complicates development. The environment should be chosen automatically during registration.

docs/PowerAuth-Push-Server-1.10.0.md
```diff
@@ -46,3 +46,17 @@ The unconfigured application list endpoint `POST /admin/app/unconfigured/list` u
...
+
+It is now possible to specify an APNs environment per device registration. The change is reflected in the REST API endpoints `/push/device/create` and `/push/device/create/multi` by addition of the `environment` parameter.
+
+The allowed values of the `environment` parameter are:
+- `development` - development APNs host is used for sending push messages
+- `production` - production APNs host is used for sending push messages
+
+For platforms other than APNs the parameter is not used, `null` value is allowed.
+
+For existing device registrations the `environment` is not specified, thus the global server setting is used. The environment is controlled by property `powerauth.push.service.apns.useDevelopment`. In case the property is set to `production`, delivery to `development` APNs host is not allowed for devices registered with `development` environment.
```
zcgandcomp: We have a global setting, an application setting (and now a device setting)

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -145,16 +151,39 @@ public boolean createDevice(String appId, String token, MobilePlatform platform,
...
+     * @return True if device registration was successful, false otherwise.
+     * @throws PushServerClientException In case of network, response / JSON processing, or other IO error.
+     */
+    public boolean createDevice(final CreateDeviceRequest request) throws PushServerClientException {
+        String error = CreateDeviceRequestValidator.validate(request);
+        if (error != null) {
+            throw new PushServerClientException(error);
+        }
+
+        logger.info("Calling create device service, appId: {}, token: {}, platform: {} - start", request.getAppId(), maskToken(request.getToken()), request.getPlatform());
```
zcgandcomp: Well, apart from the logging guideline.
"Calling create device service, appId: {}, token: {}, platform: {} - start"

```
 logger.info("call={}, callType={}, action: createDevice, state: initiated, appId: {}, token: {}, platform: {}, environment: {}", "/push/device/create", "POST" ,  request.getAppId(), maskToken(request.getToken()), request.getPlatform(),request.getEnvironment())
...
 logger.info("call={}, callType={}, action: createDevice, state: succeeded", "/push/device/create", "POST");
```

We should add the environment.
The succeeded state log  should not print the data again (because of tracing)

powerauth-push-server/src/main/java/io/getlime/push/service/AppRelatedPushClientCacheLoader.java
```diff
@@ -86,8 +90,12 @@ private AppRelatedPushClient createPushClient(final AppCredentialsEntity credent
         pushClient.setAppCredentials(credentials);
 
         if (credentials.getApnsPrivateKey() != null) {
-            final ApnsClient apnsClient = pushSendingWorker.prepareApnsClient(credentials);
-            pushClient.setApnsClient(apnsClient);
+            final ApnsClient apnsClientProduction = pushSendingWorker.prepareApnsClient(credentials, ApnsEnvironment.PRODUCTION);
+            pushClient.setApnsClientProduction(apnsClientProduction);
+            if (configuration.isApnsUseDevelopment()) {
```
zcgandcomp: This is actually on the credentials level too. So there should be something like:
` if (configuration.isApnsUseDevelopment() || credentials.getIosEnvironment()) {`

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -278,6 +288,22 @@ private void updateStatusAndPersist(PushMessageEntity pushMessageObject, PushMes
...
+        if (environmentDevice != null) {
+            final ApnsEnvironment envDevice = ApnsEnvironment.fromString(environmentDevice);
+            if (!configuration.isApnsUseDevelopment() && envDevice == ApnsEnvironment.DEVELOPMENT) {
+                logger.warn("Invalid configuration: server is configured with APNs in production mode, however push device is registered in development mode.");
+                return null;
+            }
+            return envDevice;
+        }
+        // Fallback in case device was registered without environment
+        if (configuration.isApnsUseDevelopment()) {
```
zcgandcomp: and we still have the config on the application level

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
romanstrobl: Ok, I will change the logging in this class.

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
romanstrobl: Added.

powerauth-push-server/src/main/java/io/getlime/push/service/AppRelatedPushClientCacheLoader.java
romanstrobl: Added.

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
```diff
@@ -138,23 +144,48 @@ public boolean createDevice(String appId, String token, MobilePlatform platform,
...
+        String error = CreateDeviceRequestValidator.validate(request);
+        if (error != null) {
+            throw new PushServerClientException(error);
+        }
+
+        logger.info("call={}, callType={}, action: createDevice, state: initiated, appId: {}, token: {}, platform: {}, environment: {}",
+                "/push/device/create", "POST", request.getAppId(), maskToken(request.getToken()), request.getPlatform(), request.getEnvironment());
         Response response = postObjectImpl("/push/device/create", new ObjectRequest<>(request));
-        logger.info("Calling create device service, appId: {}, token: {}, platform: {} - finish", appId, maskToken(token), platform);
+        logger.info("call={}, callType={}, action: createDevice, state: succeeded", "/push/device/create", "POST");
```
banterCZ, line 171: 🤔 Is parametrizing the log message by the static string value needed?

banterCZ (approved):
Looks good.

powerauth-push-client/src/main/java/io/getlime/push/client/PushServerClient.java
romanstrobl: I am following the previous feedback in this PR.

powerauth-push-model/src/main/java/io/getlime/push/model/validator/CreateDeviceRequestValidator.java
```diff
@@ -71,6 +72,9 @@ public static String validate(CreateDeviceForActivationsRequest request) {
         if (request.getPlatform() == null) {
             return "Platform must not be null.";
         }
+        if (request.getPlatform() != MobilePlatform.APNS && request.getEnvironment() != null) {
+            return "Environment specified for platform which does not support environment setting.";
```
zcgandcomp:
```suggestion
            return "Environment specified for a platform that does not support environment setting.";
```

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -194,14 +201,22 @@ public void sendCampaignMessage(String appId, Platform platform, String token, P
...
         final AppRelatedPushClient pushClient = prepareClients(appId);
 
         final PushMessageEntity pushMessageObject = pushMessageDAO.storePushMessageObject(pushMessageBody, attributes, userId, activationId, deviceId);
 
         switch (platform) {
-            case IOS, APNS ->
-                    pushSendingWorker.sendMessageToApns(pushClient.getApnsClient(), pushMessageBody, attributes, priority, token, pushClient.getAppCredentials().getApnsBundle(), createPushSendingCallback(token, pushMessageObject, pushClient));
+            case IOS, APNS -> {
+                final String environmentAppConfig = pushClient.getAppCredentials().getApnsEnvironment();
+                final ApnsEnvironment apnsEnvironment = environment != null ? resolveApnsEnvironment(environment.getKey(), environmentAppConfig) : resolveApnsEnvironment(null, environmentAppConfig);
```
zcgandcomp: The environment passed as an attribute is actually the device environment so why do we not use just:

final ApnsEnvironment environment =  resolveApnsEnvironment(environment, environmentAppConfig);

?

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -278,6 +293,29 @@ private void updateStatusAndPersist(PushMessageEntity pushMessageObject, PushMes
         }
     }
 
+    private ApnsEnvironment resolveApnsEnvironment(final String environmentDevice, final String environmentAppConfig) {
+        if (environmentDevice != null) {
```
zcgandcomp: Just as a code concept for explaining the resolution logic - the switching between an enum and a String is hard to follow. We should not use the enum on the Service layer at all IMHO. @banterCZ 

// resolve environment from the application configuration
String environment = environmentAppConfig;
// resolve environment from the server configuration if not set on application level
if (environment == null) {
 environment = configuration.isApnsUseDevelopment();
}
// now the environment should be already set - otherwise we have invalid configuration, but let's skip it.

// the environment is set on the device level
if (environmentDevice != null) {
// the device is set for production or the server / app is configured for development
 if (environmentDevice == ApnsEnvironment.PRODUCTION || environment == ApnsEnvironment.DEVELOPMENT) {
    // we can do the validation in this way
    environment = environmentDevice;
  } 
}

return environment;

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -278,6 +293,29 @@ private void updateStatusAndPersist(PushMessageEntity pushMessageObject, PushMes
...
+            final ApnsEnvironment envDevice = ApnsEnvironment.fromString(environmentDevice);
+            if (!configuration.isApnsUseDevelopment() && envDevice == ApnsEnvironment.DEVELOPMENT) {
+                logger.warn("Invalid configuration: server is configured with APNs in production mode, however push device is registered in development mode.");
```
zcgandcomp: The logging is excessive either log error after the call of the resolveEnvironment, or log warn in the resolveEnvironment.

zcgandcomp (changes requested):

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
romanstrobl: I will cleanup the code.

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
romanstrobl: Ok, I will use this logic. Based on test results it is identical to original code, but it seems more readable.

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
romanstrobl: Ok, I will move the code outside of this method.

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
```diff
@@ -96,13 +98,17 @@ public BasePushMessageSendResult sendPushMessage(final String appId, final Mode
 
                     final Platform platform = device.getPlatform();
                     if (platform == Platform.IOS || platform == Platform.APNS) {
-                        if (pushClient.getApnsClient() == null) {
-                            logger.error("Push message cannot be sent to APNS because APNS is not configured in push server.");
+                        final PushMessageSendResult.PlatformResult platformResult = sendResult.getApns();
+                        final PushSendingCallback callback = createPushSendingCallback(mode, device, platformResult, pushMessageObject, phaser);
+                        final String environment = resolveApnsEnvironment(device.getEnvironment(), appCredentials.getApnsEnvironment());
+                        if (!configuration.isApnsUseDevelopment() && ApnsEnvironment.DEVELOPMENT.getKey().equals(device.getEnvironment())) {
```
zcgandcomp: This will not take into the account server appCredentials setting.

we can simply as:
```
// if the device environment is used, and the device environment is not the same as resolved environment we have to skip.
if (device.getEnvironment() !=null && !device.getEnvironment().equals(environment))
```

Alternatively you can modify resolveEnvironment method and ensure null is returned in the case of incorrect state.

Example in resolveEnvironment:

```
        if (environmentDevice != null) {
            if (environmentDevice.equals(ApnsEnvironment.PRODUCTION.getKey()) || environment.equals(ApnsEnvironment.DEVELOPMENT.getKey())) {
                environment = environmentDevice;
            } else {
               // environment cannot be resolved as we have conflicting server settings and device settings
               environment = null;
           }
        }
```

zcgandcomp (changes requested):

powerauth-push-server/src/main/java/io/getlime/push/service/PushMessageSenderService.java
romanstrobl: Understood. I prefer the variant with returning `null` value and handling this state outside the method.

romanstrobl: I added a test for this case. Both global configuration and application credential configuration setting is tested.

zcgandcomp (approved):
Nice, thank you!

PR #911, opened 2024-11-19
Fix #908: The tracing context is not propagated to PushSendingWorker

### LINKED ISSUE #908: The tracing context is not propagated to PushSendingWorker
When a push notification is sent via worker, we lose the tracing context. This makes it extremely difficult to trace problems with the actual Push sending. There are multiple issues with push sending, typically due to a wrong configuration.

Propagete context to Phaser.

banterCZ:


romanstrobl (approved):
Nice solution!

PR #912, opened 2024-11-19
Fix #910: Make CA certificate loading fail early

### LINKED ISSUE #910: Make CA certificate loading fail early
We faced several regression bugs while loading the certificates. Make loading eager to fail early.

A follow-up to #909

banterCZ:


powerauth-push-server/src/main/java/io/getlime/push/service/CaCertificateService.java
```diff
@@ -30,19 +31,17 @@
...
+ * Service for working with CA Certificates.
  *
  * @author Petr Dvorak, petr@wultra.com
  */
 @Service
-@AllArgsConstructor
+@RequiredArgsConstructor
 @Slf4j
-public class CaCertUtil {
+public class CaCertificateService {
```
banterCZ, line 44: Misleading, it is not a util class.

powerauth-push-server/src/main/java/io/getlime/push/service/CaCertificateService.java
```diff
@@ -59,12 +58,21 @@ public class CaCertUtil {
...
 
+    private List<X509Certificate> allCertificates;
+
     /**
      * Obtain all registered CA certificates.
      * @return All registered CA certificates.
      */
     public X509Certificate[] allCerts() {
-        // Prepare result list
+        return allCertificates.toArray(X509Certificate[]::new);
```
banterCZ, line 68: Besides early failure, we get the performance improvement to load it just once.

romanstrobl (approved):
Nice improvement!

powerauth-push-server/src/main/java/io/getlime/push/service/CaCertificateService.java
```diff
@@ -59,12 +58,21 @@ public class CaCertUtil {
...
+    }
+
+    /**
+     * Load the locally stored CA certificates required by Apple for APNs.
+     */
+    @PostConstruct
```
banterCZ, line 74: @zcgandcomp This may affect startup time, but the benefit is, that it is loaded once per push-server instance and fails early if any error. This used to be called per APNS client and failed when called.

powerauth-push-server/src/main/java/io/getlime/push/service/CaCertificateService.java
```diff
@@ -83,29 +94,35 @@ public X509Certificate[] allCerts() {
...
-                    final String certString = new String(inputStream.readAllBytes(), StandardCharsets.UTF_8);
-                    final X509Certificate cert = certificateFromPem(certString);
-                    result.add(cert);
-                }
-            } catch (CertificateException | IOException e) {
-                logger.error("Certificate error: {}", e.getMessage(), e);
+        result.addAll(loadEmbeddedCertificate());
+
+        allCertificates = result;
+        logger.debug("#loadAllCertificates took {} ms", System.currentTimeMillis() - startTimestamp);
```
banterCZ:
```log
#loadAllCertificates took 24 ms
```

banterCZ:
As confirmed with @zcgandcomp at technical-grooming, this change does not heavily affect the start time.

PR #917, opened 2024-11-25
Fix #913: Push messages sent via FCM to APNs are not constructed correctly

### LINKED ISSUE #913: Push messages sent via FCM to APNs are not constructed correctly
We should update the payload of push messages which are sent via FCM to APNs, so that all push message parameters for APNs are mapped correctly including silentness of the push message, title, body, extras, sound, and other parameters which can be controlled using Push Server REST API.

romanstrobl:
The APNs payload is constructed and set for FCM in format expected by FCM. FCM should be able to decide which payload to use based on token value, so both Android and APNs push messages are constructed properly including all details in the platform-dependent sections of the payload.

### DIFF
--- a/powerauth-push-server/src/main/java/io/getlime/push/service/ApnsPayloadBuilder.java
+++ b/powerauth-push-server/src/main/java/io/getlime/push/service/ApnsPayloadBuilder.java
@@ -78,7 +78,7 @@ final class ApnsPayloadBuilder {
         return payloadBuilder.build();
     }
 
-    static ApnsConfig buildApnsPayloadForFcm(final PushMessageBody push, final boolean isSilent, final DeliveryPriority priority) {
+    static ApnsConfig buildPayloadForFcm(final PushMessageBody push, final boolean isSilent, final DeliveryPriority priority) {
         Aps.Builder apsBuilder = Aps.builder();
 
         if (!isSilent) {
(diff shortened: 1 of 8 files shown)


powerauth-push-server/src/main/java/io/getlime/push/service/ApnsPayloadBuilder.java
```diff
@@ -0,0 +1,131 @@
...
+ * Convert {@link PushMessageBody} to APNs platform-dependent payload.
+ *
+ * @author Lubos Racansky, lubos.racansky@wultra.com
+ * @author Roman Strobl, roman.strobl@wultra.com
+ */
+@Slf4j
+final class ApnsPayloadBuilder {
+
+    private ApnsPayloadBuilder() {
+        throw new IllegalStateException("Should not be instantiated.");
```
banterCZ: I know that I am the author of this class 😅 But today, I think that _builder_ is not a perfect name. I mean I am surprised by the hidden constructor, I would expect that `ApnsPayloadBuilder` builds `ApnsPayload` like `...builder().build()`, but this class is more a convertor from `ApnsPayloadBuilder` to APNs JSON payload, isn't it?

banterCZ (approved):
The draft looks good, I suggest testing it end-to-end.

powerauth-push-server/src/main/java/io/getlime/push/service/ApnsPayloadBuilder.java
romanstrobl: I will rename the class.

banterCZ: Thanks!

romanstrobl:
We have tested that delivery of push messages through FCM to iOS works correctly now, including handling of silentness.

PR #922, opened 2024-12-11
Fix #920: Embed new APNS certificates

### LINKED ISSUE #920: Embed new APNS certificates
As per https://developer.apple.com/news/?id=09za8wzy
we need to add new certificates to trust store.

romanstrobl:


banterCZ (approved):
Looks good.

powerauth-push-server/src/main/resources/cacert/USERTrustRSAAAACA_2025.pem
banterCZ, line 1: I have checked, that this PEM is `SHA-2 Root : USERTrust RSA Certification Authority` from https://www.sectigo.com/knowledge-base/detail/Sectigo-Intermediate-Certificates/kA01N000000rfBO as stated in https://developer.apple.com/news/?id=09za8wzy

PR #926, opened 2024-12-12
Fix #923: Creating device registration in multi-user mobile app setup overwrites the previous registration

### LINKED ISSUE #923: Creating device registration in multi-user mobile app setup overwrites the previous registration
## Use case scenario

- PowerAuth-based app that allows multiple users (multiple powerauth activations - imagine Instagram or Facebook multiple account setup)
- User 1 enrolls the device to his account (creates PowerAuth activation)
- User 1 registers for the push notifications
- User 2 enrolls the same device to his account (creates PowerAuth activation)
- User 2 registers for the push notifications
- Now only user 2 is able to receive notifications
- When user 1 registers again, only user 1 is able to receive push notifications
- etc...

## What is the problem

The problem is that when seconds activation is registered with the same push token (push token is the same for the application) as another already created record in the database, the same row is reused, even if the activationId is different.

```
# user 1 registered
1604  "e350450e-3fac-420d-b0f5-df9a43fe8666"  "koberpushtest1"  1 "ios" "D393968A5D41EECA112D348D0D34888D6AC6F84C04290347FBA3EA470852109E"  "2024-12-11 13:07:14.287" true  "development"
# User 2 registered (the same row with the ID 1604 is reused making the previous activation unable to receive notifications)
1604  "2ffb43a0-8044-4c37-8f01-c3b69889e2f0"  "koberpushtest2"  1 "ios" "D393968A5D41EECA112D348D0D34888D6AC6F84C04290347FBA3EA470852109E"  "2024-12-11 13:08:56.299" true  "development"
```

## The code

The issue is happening here, when a device is registering for push, a lookup is made:
https://github.com/wultra/powerauth-push-server/blob/develop/powerauth-push-server/src/main/java/io/getlime/push/service/DeviceRegistrationService.java#L71C60-L71C85

The lookup returns a single result (because it's matched by the push token) from here:
https://github.com/wultra/powerauth-push-server/blob/develop/powerauth-push-server/src/main/java/io/getlime/push/service/DeviceRegistrationService.java#L207

This single record is then updated making the original activation unable to receive notifications:
https://github.com/wultra/powerauth-push-server/blob/develop/powerauth-push-server/src/main/java/io/getlime/push/service/DeviceRegistrationService.java#L78

romanstrobl:


powerauth-push-server/src/main/java/io/getlime/push/service/DeviceRegistrationService.java
```diff
@@ -69,25 +69,35 @@ public void createOrUpdateDevice(final CreateDeviceRequest requestObject, final
         final String activationId = requestObject.getActivationId();
 
         final List<PushDeviceRegistrationEntity> devices = lookupDeviceRegistrations(appId, activationId, pushToken);
-        final PushDeviceRegistrationEntity device;
+        PushDeviceRegistrationEntity device = null;
```
banterCZ, line 71: 🤔 IMHO, it could be better to keep the _final_ variable, which of course requires some code changes.

powerauth-push-server/src/main/java/io/getlime/push/service/DeviceRegistrationService.java
```diff
@@ -69,25 +69,35 @@ public void createOrUpdateDevice(final CreateDeviceRequest requestObject, final
...
-            // Push token can be associated with multiple activations only when associated activations are enabled.
-            // Push device registration must be done using /push/device/create/multi endpoint in this case.
-            logger.info("Multiple device registrations found: app ID: {}, activation ID: {}, platform: {}", requestObject.getAppId(), requestObject.getActivationId(), platform);
-            throw new PushServerException("Multiple device registrations found for push token. Use the /push/device/create/multi endpoint for this scenario.");
+            // 3. A row with different activation ID but same push token is created. The user removed an activation and created a new one, the push token remains the same,
+            //    or the user registered multiple activations with the same push token.
+            for (PushDeviceRegistrationEntity existingDevice: devices) {
+                if (existingDevice.getActivationId().equals(activationId)) {
+                    device = existingDevice;
+                }
```
banterCZ: What about?
```suggestion
                    device = existingDevice;
                    logger.debug("Found existing device ID: {} for activation ID: {}", existingDevice.getId(), activationId);
                }
```

banterCZ (approved):
The code looks good, but from the product point of view, not sure whether to support this scenario and make it so complicated.

powerauth-push-server/src/main/java/io/getlime/push/service/DeviceRegistrationService.java
romanstrobl: I do not know how to achieve this easily.

petrdvorak (changes requested):
What happens if registration is removed locally on the device and the device is personalized to other user ID? Will sending push to the old user ID still arrive?

romanstrobl:
Closing pull request, we will implement this feature using a new specific endpoint.

PR #961, opened 2025-04-22
Fix #960: Improve logging for sendPushMessageBatch

### LINKED ISSUE #960: Improve logging for sendPushMessageBatch
### Task specification

When sendPushMessageBatch is called, success is returned even though some messages failed. Log summary numbers from `PushMessageSendResult.PlatformResult`.

banterCZ:


jnpsk (approved):
looks ok

zcgandcomp (changes requested):
Please follow the logging guideline.

powerauth-push-server/src/main/java/com/wultra/push/controller/rest/PushMessageController.java
```diff
@@ -126,7 +126,7 @@ public ObjectResponse<BasePushMessageSendResult> sendPushMessageBatch(@RequestBo
         final Mode mode = requestObject.getMode();
         final List<PushMessage> batch = requestObject.getBatch();
         final BasePushMessageSendResult result = pushMessageSenderService.sendPushMessage(appId, mode, batch);
-        logger.info("The sendPushMessageBatch request succeeded, application ID: {}", requestObject.getAppId());
+        logger.info("The sendPushMessageBatch request succeeded, application ID: {}, {}", requestObject.getAppId(), result);
```
zcgandcomp: We should start logging with:

  logger.info("action: {}, state: {}, applicationId: {}", "sendPushMessageBatch", "initiated", requestObject.getAppId());

And finish with:

  logger.info("action: {}, state: {}, result: {}", "sendPushMessageBatch", "succeeded", result );

I honestly have a doubts regarding the logging of the whole result object, but it is acceptable, as we want to log the job result, not excessive details and the object is reasonably small.

zcgandcomp (changes requested):

powerauth-push-server/src/main/java/com/wultra/push/controller/rest/PushMessageController.java
banterCZ: Ok, I will change it to a new form. I am touching this because we may log _succeeded_, but the result is `(sent=0, failed=1, pending=0, total=1)`, which is not visible right now.

zcgandcomp (approved):
Ok, understand the need to log more relevant information. Let's do all changes in a structured way for better observability.

PR #967, opened 2025-05-21
Fix #965: Optimize device registration

### LINKED ISSUE #965: Optimize device registration
### Task specification

When a device is registered, we call the PowerAuth Server to verify its status. This is not necessary when we know the status—e.g. when the call comes from the Enrollment Server.

Modify the Push Server API to enable send activation status as an optional parameter and skip the check if the parameter is used.

Update the migration guide and the API doc.

Test push device registration works for active devices and doesn't register non active devices.

jnpsk:


docs/PowerAuth-Push-Server-2.0.0.md
```diff
@@ -62,3 +64,10 @@ For existing device registrations if the `environment` value is not specified du
 The global setting is controlled by property `powerauth.push.service.apns.useDevelopment`. In case the property is set to `false`, delivery to `development` APNs host is not allowed for devices registered with the `development` environment.
 
 This change is reflected in database by addition of parameter `environment` in table `push_device_registration`.
+
+### Avoiding Redundant Activation Status Lookups
```
banterCZ, line 68: Strictly speaking, we target this to 1.10.x, when deploying 2.0.x, customers should already have it. Sorry, I know that several patches will be schizophrenic a little bit.

powerauth-push-server/src/main/java/com/wultra/push/service/DeviceRegistrationService.java
```diff
@@ -231,28 +245,51 @@ private static void updateDeviceRegistrationEntity(PushDeviceRegistrationEntity
     }
 
     /**
-     * Update activation for given device in case activation exists in PowerAuth server and it is not in REMOVED state.
-     * Otherwise fail the device registration because registration could not be associated with an activation.
+     * Fetch and update activation details associated with the given device.
+     * {@link PowerAuthClientException} is thrown in case the activation does not exist
```
banterCZ: 👀 Isn't it better to document that in `@throws`? Moreover, `PowerAuthClientException` is watched, isn't it?

powerauth-push-server/src/main/java/com/wultra/push/service/DeviceRegistrationService.java
```diff
@@ -231,28 +245,51 @@ private static void updateDeviceRegistrationEntity(PushDeviceRegistrationEntity
...
+    private void fetchAndUpdateActivationForDevice(PushDeviceRegistrationEntity device, String activationId) throws PushServerException {
         try {
             final GetActivationStatusResponse activation = powerAuthClient.getActivationStatus(activationId);
-            if (activation != null && !ActivationStatus.REMOVED.equals(activation.getActivationStatus())) {
-                device.setActivationId(activationId);
-                device.setActive(activation.getActivationStatus().equals(ActivationStatus.ACTIVE));
-                device.setUserId(activation.getUserId());
-                return;
+            if (activation == null) {
+                throw new PushServerException("Device registration failed because associated activation was not found");
```
banterCZ, line 259: What about including `activationId`?

banterCZ (commented):
Looks good to me. Let's handle the situation with versions in the documentation, please.

docs/PowerAuth-Push-Server-2.0.0.md
jnpsk: Thanks. As all the changes target 1.10.x, I will rename the file accordingly.

powerauth-push-server/src/main/java/com/wultra/push/service/DeviceRegistrationService.java
jnpsk: Content of the request is logged at the start of the processing. I think the process can then be tracked via span/track ID.

banterCZ (approved):

powerauth-push-server/src/main/java/com/wultra/push/service/DeviceRegistrationService.java
```diff
@@ -135,7 +149,7 @@ public void createOrUpdateDevices(final CreateDeviceForActivationsRequest reques
                 device.setTimestampLastRegistered(new Date());
                 device.setPlatform(convert(platform));
                 device.setEnvironment(environment != null ? environment.getKey() : null);
-                updateActivationForDevice(device, activationId);
+                fetchAndUpdateActivationForDevice(device, activationId);
```
zcgandcomp, line 152: We will not have optimization here, but this APIs is not so often used.

powerauth-push-server/src/main/java/com/wultra/push/service/PushDeviceService.java
```diff
@@ -50,8 +50,8 @@ public Response createDevice(final CreateDeviceRequest request) throws PushServe
         if (request == null) {
             throw new PushServerException("Request object must not be empty");
         }
-        logger.info("Received createDevice request, app ID: {}, activation ID: {}, token: {}, platform: {}, environment: {}", request.getAppId(),
-                request.getActivationId(), maskPushToken(request.getToken()), request.getPlatform(), request.getEnvironment());
+        logger.info("Received createDevice request, appId: {}, activationId: {}, token: {}, platform: {}, environment: {}, activationStatus: {}, userId: {}", request.getAppId(),
```
zcgandcomp: can we please format in a structured way the beginning of the log. Please mention the changed logging in the migration guideline.

zcgandcomp (changes requested):
Overall logic is fine. And scope is also fine. Minor request regarding the logging.

zcgandcomp (approved):
Looks OK

romanstrobl (approved):

PR #972, opened 2025-05-27
Fix #965: Optimize device registration (#967)

### LINKED ISSUE #965: Optimize device registration

jnpsk:
(cherry picked from commit 3ee7e52e6c5ea781b9b0f48b5d1d5976c741fd85)


banterCZ (approved):
Thanks for the cherry-pick.

PR #1002, opened 2025-07-16
Fix #997: Migration is skipped (backport)

### LINKED ISSUE #997: Migration is skipped
### Affected product(s) and version(s)

1.10.

### Environment

_No response_

### Steps to reproduce, current result, expected result

Steps:
Run migration to 1.10.
Expected result:
The push_app_credentials table should be data migrated.
Actual result:
The migration script is skipped.

### Technical Notes

There are two changeset that incorrectly evaluate conditions and skip the migration:

```
    <changeSet id="3" logicalFilePath="powerauth-push-server/1.10.x/20241029-add-new-platforms.xml" author="Roman Strobl">

    <changeSet id="4" logicalFilePath="powerauth-push-server/1.10.x/20241029-add-new-platforms.xml" author="Roman Strobl">
```
Suggested changes:
1) Do not use `changeSetExecuted` in the precondition - it is irrelevant if the changeset was run or not.
2) We can have the preconditions if the columns exist in the table (ie. the previous changeset was run or somebody applied it manually) but we should fail if the precondition is not met. Otherwise we will mark the migration as completed, and it will not run next time.
3) we need to add new migration changeset the old one is marked as run. (or force rerun)

There is a small window of possible problems if we have the following scenario:
1.10 is deployed (the configuration is not migrated)
The configuration is set via API
1.10.1 with fix is applied (this overrides the previous change via API)

But we should state in the migration that 10.0 should not be deployed (and remove it from docker hubs)

_No response_

banterCZ:
(cherry picked from commit bcd6b7072a296daa60827a0f173016352af6ae73)

### DIFF
--- a/docs/db/changelog/changesets/powerauth-push-server/1.10.x/20241029-add-new-platforms.xml
+++ b/docs/db/changelog/changesets/powerauth-push-server/1.10.x/20241029-add-new-platforms.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <databaseChangeLog xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
                    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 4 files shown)


romanstrobl (approved):
Looks OK.

PR #1060, opened 2026-01-21
Refactor configuration for push server to use environment variables and add Docker CI

filipbursik-wultra:
* Refactor configuration for push server to use environment variables and add Docker CI

* Refactor Docker setup to remove hardcoded application properties and streamline entrypoint command

* Rename logback configuration file from paps-logback.xml to push-logback.xml

* Fix env variables


korbelm (approved):
LGTM

PR #1111, opened 2026-04-14
Fix #1110: Use activation status without status blob whenever possible

### LINKED ISSUE #1110: Use activation status without status blob whenever possible
### Task specification

Improve performance by using activation status without activation status blob whenever it is possible to improve performance by avoiding unnecessary cryptographic operations.

romanstrobl:


jnpsk (approved):
👍

